PR #179, opened 2015-02-03
[Reviewer: Ellie] If no HSS, /location should only return 200 OK if subscriber actually exists

### LINKED ISSUE #178: Homestead without an HSS should only return LIR success if subscriber exists
As summary.

At present, Homestead without an HSS unconditionally returns a successful response to an LIR or UAR request.  It should only success if the subscriber actually exists.

Without this fix, it's not possible to configure tandem function without an HSS, as Sprout thinks all subscribers are local.

mirw:
Ellie,

Please can you review my change which fixes #178?  This is a prerequisite for fixing https://github.com/Metaswitch/sprout/issues/927 and together they fix tandem function.

I've live-tested, and UT-ed this.  (I've not yet UT-ed the fix to https://github.com/Metaswitch/sprout/issues/927 but I will do shortly.)

As you'll have seen, I've updated docs.

Matt


src/handlers.cpp
```diff
@@ -627,20 +627,10 @@ void ImpuLocationInfoTask::run()
   }
   else
   {
-    LOG_DEBUG("No HSS configured - fake response for server %s", _server_name.c_str());
+    LOG_DEBUG("No HSS configured - fake up response if subscriber exists");
     SAS::Event event(this->trail(), SASEvent::ICSCF_NO_HSS, 0);
```
eleanor-merry: I think the wording of this SAS event needs changing now as well

src/handlers.cpp
```diff
@@ -717,6 +707,62 @@ void ImpuLocationInfoTask::sas_log_hss_failure(int32_t result_code)
...
+  CassandraStore::Transaction* tsx =
+    new CacheTransaction(this,
+                         &ImpuLocationInfoTask::on_get_reg_data_success,
+                         &ImpuLocationInfoTask::on_get_reg_data_failure);
+  _cache->do_async(get_reg_data, tsx);
+}
+
+void ImpuLocationInfoTask::on_get_reg_data_success(CassandraStore::Operation* op)
+{
+  SAS::Event event(this->trail(), SASEvent::CACHE_GET_REG_DATA_SUCCESS, 0);
```
eleanor-merry: This doesn't have all the parameters for this SAS event

src/handlers.cpp
```diff
@@ -717,6 +707,62 @@ void ImpuLocationInfoTask::sas_log_hss_failure(int32_t result_code)
...
+    writer.String(JSON_SCSCF.c_str());
+    writer.String(_server_name.c_str());
+    writer.EndObject();
+    _req.add_content(sb.GetString());
+    send_http_reply(200);
+  }
+  else
+  {
+    LOG_DEBUG("No IMS subscription XML found for public ID %s - reject", _impu.c_str());
+    send_http_reply(404);
```
eleanor-merry: Mix of HTTPCodes and numbers

docs/homestead_api.md
```diff
@@ -107,5 +107,5 @@ Response:
 
 `{ "result-code": "2001", "mandatory-capabilties": [], "optional-capabilities": [] }`
 
-* 404 if the user cannot be found.
+* 404 if the user cannot be found (either a 5001 error from the HSS, or having no HSS configured and no record of the user).
```
eleanor-merry: (Not your change, but I've noticed while reviewing this) -> CAPABILITIES above is missing a T.

eleanor-merry: It would be nice to split out the 200 response code a bit more (e.g. the response won't contain a real HSS result code if there's no HSS)

eleanor-merry:
Looks good - some minor comments inline.

docs/homestead_api.md
graemerobertson: Capabilities is missing an i, not a t, right? Anyway, literally the hardest word to spell whilst typing.

eleanor-merry: I hadn't noticed that one! I meant the CAPABILIIES above

src/handlers.cpp
mirw: Done (in https://github.com/Metaswitch/clearwater-sas-resources).

PR #181, opened 2015-02-09
Unit test needs to link libdl when built using Ubuntu 14.04's tools chain (gcc 4.8.2)

dtheobald:
The unit tests would not run on Ubuntu 14.04 with specifically linking against libdl, so that has been added to the unit test ld flags. With this option the unit tests run as expected and has no deleterious affect in an Ubuntu 12.04 build environment.


AlexHockey:
Just to be clear - are you saying that you successfully built and run the homestead UT's on 12.04? If so then I'm happy.

dtheobald:
Yes, it successfully builds and run's the UTs on 12.04.

AlexHockey:
Great, thanks!

PR #184, opened 2015-02-19
[Reviewer: Rob] Make load monitor options configurable

eleanor-merry:
Rob, can you review these changes to make the load monitor configurable. They're basically the same as https://github.com/Metaswitch/sprout/pull/944


rkd-msw:
Looks good.

PR #193, opened 2015-03-17
Reviewer AMC: Default diameter timeout using target latency upstream

steelyvoid:
Defaults the diameter_timeout_ms config parameter to twice the target latency (target_latency_us). As the latter is expressed in microseconds and the former in milliseconds, this means dividing the value of target_latency_us by 500 (rounding up).

Tested by configuring each of the four combinations, i.e.

- diameter_timeout_ms and target_latency_us configured
- diameter_timeout_ms configured/target_latency_us not configured
- diameter_timeout_ms not configured/target_latency_us configured
- neither diameter_timeout_ms nor target_latency_us configured

and checked (via homestead DEBUG logs) that the expected values were passed to the homestead executable.


debian/homestead.init.d
```diff
@@ -122,6 +122,15 @@ get_settings()
...
         fi
 
+        # Default the diameter timeout to twice the target latency if not
+        # already overridden (rounding up).  Note that the former is expressed
+        # in milliseconds and the latter in microseconds, hence division by 500
+        # (i.e. multiplication by 2/1000).
+        if [ -z $diameter_timeout_ms ]
+        then
+          diameter_timeout_ms=$(( ($target_latency_us + 499)/500 ))
+        fi
```
bossmc: You seem to have dropped the execute bit on this script.

bossmc:
On comment on permissions, otherwise fine.

PR #194, opened 2015-03-17
[Reviewer: Ellie] Homestead depends on libevent-2.0-5 and libcurl3-gnutls

### LINKED ISSUE #192: homestead depends on libevent-2.0-5 and libcurl3-gnutls...
...but this isn't specified in debian/control.

Not seen very often because these are often dependencies of other installed components.  However, it is visible on very stripped-down cuts of Ubuntu.

mirw:
Ellie,

Please can you review this, which fixes #192.

I've tested live, and have similar changes to other components ready to go.

Matt

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -11,7 +11,7 @@ Homepage: http://projectclearwater.org/
 
 Package: homestead-libs
 Architecture: any
-Depends: libsctp1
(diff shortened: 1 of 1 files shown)


eleanor-merry:
Looks good

mirw:
Actually, it's not clear whether homestead should depend on libevent.

PR #196, opened 2015-03-22
[Reviewer: Matt] Pass the Diameter hostname to RealmManager along with the realm

rkd-msw:
This is the change we've discussed a couple of times now - to invoke RealmManager to manage the connections even if just a hss_hostname is specified, so the code is more consistent and predictable.

I'll test by setting up a HSS and configuring each of the three options:
* just hss_realm
* just hss_hostname
* both hss_realm and hss_hostname

and check that the CE/CEA exchage is successful in each case.

It's probably worth renaming RealmManager - how about DiameterConnectionManager?

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -743,21 +743,16 @@ int main(int argc, char**argv)
 
   DiameterResolver* diameter_resolver = NULL;
   RealmManager* realm_manager = NULL;
-  Diameter::Peer* peer = NULL;
-
-  if (!options.dest_realm.empty())
+  
+  if (hss_configured)
   {
     diameter_resolver = new DiameterResolver(dns_resolver, af);
     realm_manager = new RealmManager(diameter_stack,
                                      options.dest_realm,
+                                     options.dest_host,
                                      options.max_peers,
                                      diameter_resolver);
     realm_manager->start();
-  }
-  else if (!(options.dest_host.empty() || options.dest_host == "0.0.0.0"))
-  {
-    peer = new Diameter::Peer(options.dest_host);
-    diameter_stack->add(peer);
   }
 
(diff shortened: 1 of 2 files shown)


modules/cpp-common
```diff
@@ -1 +1 @@
-Subproject commit a9d63716ee1413ba59bbe987f3375f847ac84759
+Subproject commit dfb3da182d799398dac3f309bc1a403fe963cbda
```
mirw: Probably need to fix up before merging.

src/main.cpp
```diff
@@ -743,22 +743,17 @@ int main(int argc, char**argv)
...
   {
     diameter_resolver = new DiameterResolver(dns_resolver, af);
     realm_manager = new RealmManager(diameter_stack,
                                      options.dest_realm,
+                                     options.dest_host,
                                      options.max_peers,
                                      diameter_resolver);
     realm_manager->start();
   }
-  else if (!(options.dest_host.empty() || options.dest_host == "0.0.0.0"))
```
mirw: Previously, we checked for `0.0.0.0` here and didn't add it.  Will the new code still ignore `0.0.0.0` or will it end up trying to connect?  (Presumably we use that when we don't want to connect to an HSS.)

mirw:
Looks good.  A couple of comments.

src/main.cpp
rkd-msw: No, we're safe - see https://github.com/Metaswitch/homestead/blob/realmmanager_with_host/src/main.cpp#L689, where hss_configured is false of options.dest_host == "0.0.0.0".

rkd-msw:
Tested, ready for merging alongside the cpp-common change.

PR #197, opened 2015-03-22
[Reviewer: Andy] Add max_log_directory_size config option

rkd-msw:
This lets you change the max size of a log directory from the 1GB default, by setting `max_log_directory_size=<bytes>` in /etc/clearwater/config.

This would let us fix https://github.com/Metaswitch/sprout/issues/806, by setting a 50MB limit on all-in-one nodes (i.e. documenting that you should do that, making Chef do that, and setting it in https://github.com/Metaswitch/clearwater-infrastructure/blob/master/clearwater-auto-config/etc/clearwater/config for the OVF).

I'll test by setting it to something reasonably small (1KB?) and checking that most logs are deleted every hour.

(If this is OK, I'll apply it without review to all the other repositories with log-cleanup files).


bossmc:
`clearwater-auto-configure`is not only used for AIO nodes, it can also be used for generic installs.  Otherwise fine.

graemerobertson:
I'm testing this. I'm intending to port this over to our other CPP projects, and also document the new configuration option. However, I'm intending to leave https://github.com/Metaswitch/sprout/issues/806 open until at least the Chef changes are made (it doesn't sound like we could do any better than that though).

PR #198, opened 2015-03-22
[Reviewer: Ellie] Better errors when /etc/clearwater/config is missing settings

### LINKED ISSUE #140: Unhelpful errors when /etc/clearwater/config is malformed
I made a typo and had /etc/clearwater/config contain a "public_hostnamme" variable, not "public_hostname". The only symptom was this error whenever Homestead tried to create keys for securing Diameter:

```
unable to load certificate
140078750860960:error:0906D06C:PEM routines:PEM_read_bio:no start line:pem_lib.c:696:Expecting: TRUSTED CERTIFICATE
certtool: import error: The peer did not send any certificate.
certtool: import error: The peer did not send any certificate.
certtool: import error: The peer did not send any certificate.
** Note: Please use the --sec-param instead of --bits
Generating a 1024 bit RSA private key...
Generating a self signed certificate...
set_dn: The request is invalid.
```

A "public_hostname is not set" error would have been more useful.

rkd-msw:
Fixes #140. I'll test by doing a manual install with "public_hostnamme" set instead of "public_hostname" and check that the error is sensible.

### DIFF
--- a/homestead.root/usr/share/clearwater/infrastructure/scripts/homestead
+++ b/homestead.root/usr/share/clearwater/infrastructure/scripts/homestead
@@ -55,7 +55,20 @@ acl_required=false
 
 if [[ -z "$identity" ]];
 then
+    if [[ -z "$public_hostname" ]];
+    then
+        echo "ERROR: must have a public_hostname set in /etc/clearwater/config" >&2
+        exit 1
+    fi
+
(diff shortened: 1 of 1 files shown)


eleanor-merry:
Looks good, and the test sounds sensible

Issue #205, opened 2015-04-06
init.d script passes zero for --diameter-timeout-ms-arg to homestead


dtheobald:
If /etc/clearwater/config doesn't contain a setting for 'diameter_timeout_ms' and 'target_latency_us' then a zero (0) is passed for the value of --diameter-timeout-ms-arg.

I've changed the following lines in my homestead init.d script:
```
        if [ -z $diameter_timeout_ms ]
        then
           diameter_timeout_ms=$(( ($target_latency_us + 499)/500 ))
        fi
```

 to:

```
        if [ -z $diameter_timeout_ms ]
        then
            if [ ! -z $target_latency_us ]
            then
               diameter_timeout_ms=$(( ($target_latency_us + 499)/500 ))
            fi
        fi
```

PR #206, opened 2015-04-07
[Reviewer: Andy] Don't use target_latency_us if it's unset

rkd-msw:
Fixes https://github.com/Metaswitch/homestead/issues/205. I've tested by running "pgrep -lf homestead" and checking that there was no "--diameter-timeout-ms=0" in the output.


bossmc:
:+1:

PR #207, opened 2015-04-10
[Reviewer: Andy] Add service run command to run blocking

mirw:
Andy,

Please can you review my enhancement to add a `service ... run` option to run a service blocking (i.e. not as a daemon)?  Most of the changes are refactoring so that run and start can share the same code.

This has been tested live.

Thanks,

Matt

### DIFF
--- a/debian/homestead.init.d
+++ b/debian/homestead.init.d
@@ -76,6 +76,19 @@ log_directory=/var/log/$NAME
 . /lib/lsb/init-functions
 
 #
+# Function to set up environment
+#
+setup_environment()
+{
+        export LD_LIBRARY_PATH=/usr/share/clearwater/homestead/lib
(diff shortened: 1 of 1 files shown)


debian/homestead.init.d
```diff
@@ -228,6 +245,17 @@ do_stop()
 }
 
 #
+# Function that runs the daemon/service in the foreground
+#
+do_run()
+{
```
bossmc: We should probably do a `--test` first since calling `run` when the service is started is a bad idea (tm).

bossmc:
One comment, otherwise fine.

debian/homestead.init.d
mirw: As discussed, this is a bit difficult/messy because "run" services don't write a PID.  We agreed not to do this.

PR #212, opened 2015-04-22
[Reviewer: Matt] Sto1226

eleanor-merry:
Matt, can you review these changes to use the new Diameter module and plumb through the stats


src/main.cpp
```diff
@@ -577,7 +577,29 @@ int main(int argc, char**argv)
...
+    "H_diameter_invalid_dest_realm",
+  };
+
+  const static int num_known_stats = sizeof(known_stats) / sizeof(std::string);
+  LastValueCache* lvc = new LastValueCache(num_known_stats,
+                                           known_stats,
+                                           "homestead",
+                                           1000);
+  StatisticsManager* stats_manager = new StatisticsManager(lvc);
+  StatisticCounter* realm_counter =  new StatisticCounter("H_diameter_invalid_dest_realm",
```
mirw: Why aren't these members of the StatisticsManager class?

homestead.root/usr/share/clearwater/infrastructure/scripts/homestead
```diff
@@ -36,6 +36,8 @@
 
 hss_port=3868
 hs_secure_listen_port=5658
+acl_required=false
+check_destination_host=true
```
mirw: What's our plan for changing the default value of `check_destination_host` in future installs?

mirw:
A couple of comments.  Otherwise looks good.

PR #213, opened 2015-04-23
[Reviewer: Matt] Fix memory leak in UT

rkd-msw:
The Jenkins memory leak errors were because I hadn't created a Diameter::Message and checked its properties in my new UT (to minimise the amount of things each test was asserting). Unfortunately, doing that also moves the contents of `_caught_fd_msg` onto the stack, so if we don't do that we leak memory.

This just creates (but doesn't use) a Diameter::Message to make sure that `_caught_fd_msg` is cleaned up properly.


mirw:
Looks good.

PR #222, opened 2015-05-20
[Reviewer: Rob] Send 503 on local cassandra connection errors

### LINKED ISSUE #221: Requests that fail because the local cassandra isn't running aren't retried
This is because we return a 504 on cache errors (with the exception of not found errors which return 404). We should return a 503 on connection errors so that Sprout will retry requests

eleanor-merry:
Rob, can you review this fix to send back on 503 on local cassandra errors. 

Fixes #221 

Tested live

### DIFF
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -194,11 +194,21 @@ void ImpiTask::on_get_av_failure(CassandraStore::Operation* op,
     LOG_DEBUG("No cached av found for private ID %s, public ID %s - reject", _impi.c_str(), _impu.c_str());
     send_http_reply(HTTP_NOT_FOUND);
   }
+  else if (error == CassandraStore::CONNECTION_ERROR)
+  {
+    // If the cache error is a failure to connect to the local cassandra,
+    // then we want Sprout to retry the request to another Homestead (as this
+    // could be a local issue to the node). Send a 503
+    LOG_DEBUG("Cache query failed as unable to connect to local cassandra");
+    send_http_reply(HTTP_SERVER_UNAVAILABLE);
+  }
   else
   {
+    // Send a 504 in all other cases (the request won't be retried)
     LOG_DEBUG("Cache query failed with rc %d", error);
     send_http_reply(HTTP_GATEWAY_TIMEOUT);
   }
+
   delete this;
 }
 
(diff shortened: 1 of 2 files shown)


src/handlers.cpp
```diff
@@ -194,11 +194,21 @@ void ImpiTask::on_get_av_failure(CassandraStore::Operation* op,
     LOG_DEBUG("No cached av found for private ID %s, public ID %s - reject", _impi.c_str(), _impu.c_str());
     send_http_reply(HTTP_NOT_FOUND);
   }
+  else if (error == CassandraStore::CONNECTION_ERROR)
+  {
+    // If the cache error is a failure to connect to the local cassandra,
+    // then we want Sprout to retry the request to another Homestead (as this
+    // could be a local issue to the node). Send a 503
```
rkd-msw: This is really picky, but can we have a full stop after 503?

src/handlers.cpp
```diff
@@ -194,11 +194,21 @@ void ImpiTask::on_get_av_failure(CassandraStore::Operation* op,
...
+    LOG_DEBUG("Cache query failed as unable to connect to local cassandra");
```
rkd-msw: Also really picky, but can this be "Cache query failed: unable to connect to local cassandra"? "[Failure]: [reason]" is what we use elsewhere (e.g. line 1322 below) and "[Failure] as [reason]" looks a bit odd to me.

rkd-msw: More pickiness: can Cassandra have a capital C everywhere (which is how http://cassandra.apache.org/ styles it)?

src/ut/handlers_test.cpp
```diff
@@ -2618,6 +2653,41 @@ TEST_F(HandlersTest, IMSSubscriptionCacheNotFound)
   t->on_failure(&mock_op);
 }
 
+// Connection failures should translate into a 503 Service Unavailable error
+TEST_F(HandlersTest, IMSSubscriptionCacheConnectionFailure)
+{
+  // This test tests an IMS Subscription handler case where the cache
+  // has an unknown failure. Start by building the HTTP request which
```
rkd-msw: Comment is wrong - it's not an unknown failure.

src/ut/handlers_test.cpp
```diff
@@ -1782,6 +1782,41 @@ TEST_F(HandlersTest, DigestIMPUNotFound)
   t->on_failure(&mock_op);
 }
 
+TEST_F(HandlersTest, DigestNoIMPUCacheConnectionFailure)
+{
+  // This test tests an Impi Digest task case where no public ID is specified
+  // on the HTTP request, the cache request fails. Start by building the HTTP
```
rkd-msw: Worth a comment at the top to explain the difference between this and DigestNoIMPUCacheFailure?

src/ut/handlers_test.cpp
```diff
@@ -4140,6 +4237,43 @@ TEST_F(HandlerStatsTest, DigestCacheFailure)
   delete _caught_diam_tsx; _caught_diam_tsx = NULL;
 }
 
+TEST_F(HandlerStatsTest, DigestCacheConnectionFailure)
+{
+  // Test that UNsuccessful cache requests result in the latency stats being
```
rkd-msw: Shouldn't this test just ignore latency and check for a 503, so it's only testing one thing?

rkd-msw:
Code looks good, I nitpicked your comments and log text a bit.

PR #223, opened 2015-05-22
Reviewer Ellie - Don't fail infrastructure scripts

bossmc:
We can't fail infrastructure scripts just because shared config hasn't arrived yet.  Tested live.


homestead.root/usr/share/clearwater/infrastructure/scripts/homestead
```diff
@@ -68,7 +68,7 @@ fi
 if [[ -z "$hs_hostname" ]];
 then
     echo "ERROR: must have a hs_hostname set in /etc/clearwater/config" >&2
```
eleanor-merry: This shouldn't be /etc/clearwater/config anymore

eleanor-merry:
OK.

PR #229, opened 2015-06-09
[Reviewer: Rob] Add in log explicitly saying Homestead is quitting

eleanor-merry:
Adding in a extra log to explicitly show that Homestead quits when Cassandra is down


rkd-msw:
There are a couple of other `exit(2)` calls - should we make the same change there, maybe factoring out a `quit` function that calls closelog, makes that LOG_STATUS and calls exit?

PR #231, opened 2015-06-19
Reviewer MIRW: Set default blacklist durations

graemerobertson:
See https://github.com/Metaswitch/ralf/pull/160


mirw:
Looks good.

PR #237, opened 2015-07-27
[Reviewer: Rob] Add JSON file and parser

eleanor-merry:
Rob, can you please review this change to move Homestead to use JSON alarms


src/alarm_header.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+ * under which the OpenSSL Project distributes the OpenSSL toolkit software,
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+#include "json_alarms.h"
+
+int main(int argc, char**argv)
+{
+  // Parse the JSON alarms and generate a header file that has the alarm IDs
+  std::string path = "../homestead.root/usr/share/clearwater/infrastructure/alarms/homestead_alarms.json";
```
rkd-msw: Is this file going to be identical in every repository apart from the path and the word "homestead"? If so, shouldn't it be in cpp-common and take those as arguments?

src/alarm_header.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  bool rc = JSONAlarms::validate_alarms_from_json(path, result, unused, header);
+
+  if (rc)
+  {
+    JSONAlarms::write_header_file("homestead", header);  
+  }
+  else
+  { 
+    printf("Invalid JSON file. Error: %s", result.c_str());
```
rkd-msw: `fprintf(stderr, ...`?

src/alarm_header.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * as those licenses appear in the file LICENSE-OPENSSL.
+ */
+#include "json_alarms.h"
+
+int main(int argc, char**argv)
+{
+  // Parse the JSON alarms and generate a header file that has the alarm IDs
+  std::string path = "../homestead.root/usr/share/clearwater/infrastructure/alarms/homestead_alarms.json";
+  std::string result;
+  std::vector<AlarmDef::AlarmDefinition> unused;
```
rkd-msw: Why is this unused (or when would you use it)?

rkd-msw:
Looks good. I'm slightly cautious about copy-pasting the alarms C++ file and Makefile into every repository, if either could just live in cpp-common.

PR #245, opened 2015-08-13
[reviewer: Rob] New stats architecture

alango:


include/statisticsmanager.h
```diff
@@ -38,50 +38,32 @@
...
-  virtual void update_##NAME(unsigned long sample) { (NAME).accumulate(sample); }
-
 class StatisticsManager : public HttpStack::StatsInterface
 {
 public:
-  StatisticsManager(LastValueCache* lvc);
+  StatisticsManager();
   virtual ~StatisticsManager();
 
-  ACCUMULATOR_UPDATE_METHOD(H_latency_us);
```
rkd-msw: It feels like this change might be simpler if you'd kept these ACCUMULATOR_UPDATE_METHOD method calls - I think that would mean that the change was limited to StatisticsManager, and it would expose the same `update_H_hss_latency_us` methods to everything else (so the existing UTs, for example, would work unchanged).

src/ut/fakestatisticsmanager.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include "fakestatisticsmanager.hpp"
+
+namespace SNMP
+{
+class FakeAccumulatorTable: public AccumulatorTable
+{
+  public:
+    int _count;
+    FakeAccumulatorTable() { _count = 0; };
+    void accumulate(uint32_t sample) { _count += sample; };
```
rkd-msw: This might not be a problem if you make the ACCUMULATOR_UPDATE_METHOD change I mentioned above, but the use of `_count` here is misleading - it's actually the sum of all samples, not the count of samples (as it is in a normal AccumulatorTable).

src/ut/handlers_test.cpp
```diff
@@ -4262,7 +4256,7 @@ TEST_F(HandlerStatsTest, DigestCacheConnectionFailure)
 
   // Cache latency stats are updated when the transaction fails.
   EXPECT_CALL(*_httpstack, send_reply(_, 503,  _));
-  EXPECT_CALL(*_stats, update_H_cache_latency_us(_));
+  ////!!EXPECT_EQ(_, ((SNMP::FakeAccumulatorTable*)_stats->H_cache_latency_us)->_count);
```
rkd-msw: Worth deleting this (or replacing with a greater-than-zero expectation)?

rkd-msw:
A couple of minor tweaks, and one place where I think restoring some of the original code might simplify this a lot.

PR #253, opened 2015-09-07
[reviewer: Rob] Cx stats

alango:


include/handlers.h
```diff
@@ -144,10 +147,12 @@ class HssCacheTask : public HttpStackUtils::Task
     StatsFlags _stat_updates;
     response_clbk_t _response_clbk;
     timeout_clbk_t _timeout_clbk;
+    SNMP::CxCounterTable* _cx_results_tbl;
 
     void on_timeout()
     {
       update_latency_stats();
+      _cx_results_tbl->increment(SNMP::DiameterAppId::TIMEOUT, 0);
```
rkd-msw: Worth explaining why the '0' is here?

src/handlers.cpp
```diff
@@ -1713,6 +1743,7 @@ void RegistrationTerminationTask::get_assoc_primary_public_ids_failure(Cassandra
   SAS::Event event(this->trail(), SASEvent::DEREG_SUCCESS, 0);
```
rkd-msw: I think this should be DEREG_FAIL instead of DEREG_SUCCESS - do you agree? (I realise this isn't your code.)

src/handlers.cpp
```diff
@@ -1670,6 +1698,7 @@ void RegistrationTerminationTask::run()
     SAS::Event event(this->trail(), SASEvent::INVALID_DEREG_REASON, 0);
     SAS::report_event(event);
     send_rta(DIAMETER_REQ_FAILURE);
+    rtr_results_tbl->increment(SNMP::DiameterAppId::BASE, 5012);
```
rkd-msw: It's a bit of a shame that we have the string "DIAMETER_UNABLE_TO_COMPLY" and then have to separately have the constant 5012. One way to make this a bit nicer would be to move this increment inside send_rta() - it can then check its string argument and update the table appropriately.

rkd-msw: (Same for send_ppa.)

rkd-msw:
Looks good - my main concern is about having to have the code manually map between result code strings and their numeric values, and whether we can move all that mapping into a common place.

src/handlers.cpp
alango: Yes, I'll make the change.

PR #259, opened 2015-10-02
[Reviewer: Ellie] Add force_hss_peer config option

rkd-msw:
Frequently (particularly in environments where DNS takes a long time to change - e.g. another team is managing the DNS server), I've wanted to be able to say "the HSS is reachable as IP address XXX, but the Destination-Host should be YYY, and the Destination-Realm should be ZZZ". (Currently, if you set hss_realm or hss_hostname, that gets resolved to find the peers, so there's no good way to uncouple the Diameter AVPs from the actual connection.)

This adds a config option to make that possible, so you can say:

```
force_hss_peer=XXX
hss_hostname=YYY
hss_realm=ZZZ
```

I haven't yet tested this live, but will do so.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -72,6 +72,7 @@ struct options
   std::string cassandra;
   std::string dest_realm;
   std::string dest_host;
+  std::string force_hss_peer;
   int max_peers;
   std::string server_name;
   int impu_cache_ttl;
@@ -115,7 +116,8 @@ enum OptionTypes
   MIN_TOKEN_RATE,
   EXCEPTION_MAX_TTL,
   HTTP_BLACKLIST_DURATION,
-  DIAMETER_BLACKLIST_DURATION
+  DIAMETER_BLACKLIST_DURATION,
+  FORCE_HSS_PEER
 };
 
 const static struct option long_opt[] =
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -811,11 +820,22 @@ int main(int argc, char**argv)
...
+      realm_manager = new RealmManager(diameter_stack,
+                                       options.dest_realm,
+                                       options.dest_host,
+                                       options.max_peers,
+                                       diameter_resolver);
+    }
+    else
+    {
+      realm_manager = new RealmManager(diameter_stack,
+                                       "",
```
eleanor-merry: Don't we still need the realm here?

eleanor-merry: You think this function is provided by the diameter.conf file - you're going to include this in your testing

eleanor-merry:
LGTM

src/main.cpp
rkd-msw: Where 'this function' is 'checking that on a CER/CEA exchange the peer has the right realm'.

rkd-msw: This doesn't happen - we've agreed to just document this restriction.

PR #260, opened 2015-10-02
[Reviewer: Andy] Allow configuring the Cx MAR authentication scheme more flexibly

rkd-msw:
I've encountered HSSes that, when you send them an authentication scheme of "Unknown", send back weird things like "Early IMS Security".

This adds two new config options to counter that:

`hss_mar_force_digest=Y` - that AVP will always be 'SIP Digest'
`hss_mar_force_aka=Y` - that AVP will always be 'Digest-AKAv1-MD5'

To allow the quotes in "SIP Digest", I've used 'eval', as recommended by http://stackoverflow.com/questions/7454526/bash-variable-containing-multiple-args-with-quotes/21163341#21163341.

I've tested live by setting this config and checking the log at https://github.com/Metaswitch/homestead/blob/dev/src/main.cpp#L332.

### DIFF
--- a/debian/homestead.init.d
+++ b/debian/homestead.init.d
@@ -154,8 +154,8 @@ get_daemon_args()
         fi
 
         [ "$hss_mar_lowercase_unknown" != "Y" ] || scheme_args="--scheme-unknown=unknown"
-        [ "$hss_mar_force_digest" != "Y" ] || scheme_args="--scheme-unknown=\"SIP Digest\" --scheme-aka=\"SIP Digest\""
-        [ "$hss_mar_force_aka" != "Y" ] || scheme_args="--scheme-unknown=Digest-AKAv1-MD5 --scheme-digest=Digest-AKAv1-MD5"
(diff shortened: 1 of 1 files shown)


debian/homestead.init.d
```diff
@@ -153,7 +153,9 @@ get_daemon_args()
           diameter_timeout_ms=$(( ($target_latency_us + 499)/500 ))
         fi
 
-        [ "$hss_mar_lowercase_unknown" != "Y" ] || scheme_unknown_arg="--scheme-unknown=unknown"
+        [ "$hss_mar_lowercase_unknown" != "Y" ] || scheme_args="--scheme-unknown=unknown"
+        [ "$hss_mar_force_digest" != "Y" ] || scheme_args="--scheme-unknown=\"SIP Digest\" --scheme-aka=\"SIP Digest\""
+        [ "$hss_mar_force_aka" != "Y" ] || scheme_args="--scheme-unknown=Digest-AKAv1-MD5 --scheme-digest=Digest-AKAv1-MD5"
```
bossmc: Also pass the other `--scheme-X` (otherwise we're relying on the default matching the hard-coded values here)?

bossmc:
Ok

PR #264, opened 2015-10-08
[Reviewer: Ellie] Add quotes around server_name to cope with semicolons

rkd-msw:
I've tested live, and confirmed:
* `ps aux | grep homestead` gives me the full command line (not cut off at the first semicolon)
* `sudo service homestead restart` doesn't give the error `/etc/init.d/homestead: 1: eval: --impu-cache-ttl=0: not found`


eleanor-merry:
LGTM

Issue #269, opened 2015-10-20
scheme setup scripts always wait for the grace period during install


dtheobald:
On initial install, the homestead infrastructure script always waits the grace period since cassandra has not been started.

mirw:
OK, so the issue here is that we install a load of packages all in one go, but dpkg breaks this into a series of stages.

*   First, it unpacks the packages onto the disk - this writes files into `/usr/share/clearwater/infrastructure/scripts`, including scripts for homestead and homer.

*   Then, it spins through all the packages, configuring them and running their `postinst` scripts, and then their `init.d` scripts.

The `clearwater-infrastructure` package's `init.d` script calls into all the scripts in the `/usr/share/clearwater/infrastructure/scripts`.  The `homestead` script installs the schema.  To do this it expects Cassandra to be running, and if it isn't it waits for a while for it to start.

The fix is probably for `/usr/share/clearwater/infrastructure/scripts/homestead` (and similarly for homer and memento) to check whether their package is installed (or at least configured) before trying to install the schema.

While investigating this, I've also noticed that our chef and AIO install processes don't match the documented manual install procedure.  I'll raise separate issues to get that fixed up.

mirw:
@JimBunch, I'm going to suggest this isn't critical.  Do you agree?

JimBunch:
Yes, not critical.

mirw:
Fixed by the above commit.

eleanor-merry:
Reopening as we've seen this happen on nodes which have the fix. I think it would be better to have the checks in the schema itself as there's multiple different ways of calling the schema script

rkd-msw:
Another possible improvement here is to declare `Pre-Depends: clearwater-cassandra` in Homestead and Crest's debian/control files. This means that Homestead won't be unpacked (including unpacking things into `/usr/share/clearwater/infrastructure/scripts/`) until clearwater-cassandra is fully installed (and so Cassandra is on its way to starting up).

rkd-msw:
Although we've been concentrating on not installing schemas until Cassandra is **configured**, https://github.com/Metaswitch/clearwater-monit/issues/22 might also exacerbate the problem, by causing Cassandra not to be **started** for a while.

rkd-msw:
Fixed by https://github.com/Metaswitch/crest/issues/269, https://github.com/Metaswitch/clearwater-cassandra/pull/75, etc.

PR #270, opened 2015-10-20
[Reviewer: Ellie] Check that Cassandra is installed before installing schema

mirw:
Ellie,

Please can you review this fix to #269?  There'll be similar fixes coming to crest and memento shortly.  Basically, we just check that Cassandra is installed (which means it should be running) before we installed the schema.

Tested live.

Matt


eleanor-merry:
LGTM

Issue #268, opened 2015-10-19
Isolating one Homestead VM of a duplex pair from the signalling network causes call failures for 20 seconds.


plfrancis:
While running a 25 CPS traffic load, isolated a Homestead VM from the signalling network. This resulted in 500 call failures over the next 20 seconds. The cause was Sprout timing out waiting for registration state and IFCs from Homestead. Homestead in turn was waiting on Cassandra. For the first number of failures Homestead simply reported the following after 14 seconds:

```
Failed to get registration information from the cache - reject the request
```

After this, Homestead started reporting the following after 18 seconds:

```
Insufficient nodes to read Cassandra at consistency level LOCAL_QUORUM
Insufficient nodes to read Cassandra at consistency level QUORUM
```

Subsequent errors were of the previous variety with the time between request and Cassandra error response decreasing to 0 over the next 18 seconds. At that point calls were once again processed normally. Diagnostics will be uploaded.

mirw:
It looks as though `LOCAL_QUORUM` and `QUORUM` reads can be very slow in this window, and this means that we hold up the `ONE` read for a long time (until the original HTTP request has long timed out).

I suspect this has been introduced by https://github.com/Metaswitch/cpp-common/commit/bf130d194058c1ed583131db80b48bfff9352d9f.

eleanor-merry:
Hi @plfrancis - passing this back to you to determine if taking the fix is worth while.

plfrancis:
Hi @eleanor-merry - with the fix it looks like Homestead no longer waits for Cassandra for long periods. But, calls still fail for 20 seconds after a Homestead VM is isolated from the network. The new issue seems to be that Sprout and Homestead are both using a 1/2 second timer and Sprout times out first before Homestead can return its 504 response (just milliseconds later). I've uploaded a SAS trace showing the new scenario.

plfrancis:
For some reason I was thinking if Sprout received the response from Homestead before timing out it would behaved differently. But, for this scenario that is probably not the case. The real question I guess is why Homestead timed out reading the cache when one Cassandra instance is still available.

AlexHockey:
A couple of questions for you @plfrancis:

* Is your deployment single-site or GR?
* Can you paste your current cassandra.yaml into the issue (there is one in the diags bundle you uploaded in October but I want to make sure I'm looking at the latest version).

AlexHockey:
[...]

PR #271, opened 2015-10-22
[Reviewer: Alex] Update UTs and schema

eleanor-merry:
Update UTs to handle TWO reads (from https://github.com/Metaswitch/cpp-common/pull/390), and update the schema to set the speculative retry value. 

If this is OK I'll make the same change to the other cassandra schemas, update the SAS bundle and add speculative_retry_value to the documentation

Part of the fix needed for #268


AlexHockey:
Looks good.

PR #284, opened 2015-11-27
[Reviewer: Andy] Add known benign leak to homestead exceptions

steelyvoid:
This is the fix to the homestead UTs which complain (on some build systems) about an apparent memory leak in gnutls_init (as called by fd_core_initialize during DiameterStack startup).  

This is benign.  While the memory is indeed apparently leaked during the test, Homestead does not repeatedly start Diameter stacks outside of the UTs, so the leak can be ignored.


bossmc:
Thanks

PR #286, opened 2015-12-04
[Reviewer: Andy] Catch a thrift timeout exception and retry when doing a HA get

AlexHockey:
Unit test for https://github.com/Metaswitch/cpp-common/pull/417


modules/cpp-common
```diff
@@ -1 +1 @@
-Subproject commit 62b457037b50e7f6e033223b485d9538bd1c9928
+Subproject commit 46bf9f809f0807e8114b3440af908860fcb2db09
```
bossmc: Needs updating.

bossmc:
Fine.

PR #291, opened 2015-12-23
[Reviewer AMC] Wait for cassandra writes to complete

steelyvoid:
Previously, homestead did not wait for cassandra writes to complete before responding to sprout->homestead requests.  One consequence of this is that an attempt to query registration state immediately after a PUT of reg-state=reg (such as that issued as part of a reg-event SUBSCRIBE immediately following a REGISTER) could sometimes overtake the put of the new state, and therefore fail to find the new registration state in cache, causing the higher level SIP verb to fail unexpectedly.

This fix uses the callbacks on the cache interface to explicitly wait for writes to the various tables so that any queries on those tables performed during the client's next operation don't encounter out of date data (exception: the PutAssociatedPrivateID operation performed as part of a fetch of the IMS subscription data from cache is not waited on, as this would involve substantial restructuring of the handler code and its not clear that there are any back to back client operations that would be affected by asynchronous update here.  This omission is tracked in an issue)

All success and failure cases are covered by UTs and the success paths have been verified in FV testing.  Some notes

-  The failure paths have not been FVed and so there has been no testing that the failure path SAS logs are correctly generated
-  The end-user consequences of a write error path are as follows
    -  If an attempt to put registration data to the cache fails, this manifests as a 500 response on the reg-data PUT command
    -  If an attempt to put associated public IDs to cache fails, this is SAS logged, but the GET of the Authentication Vector that prompted the cache write is allowed to complete successfully
    -  If an attempt to delete IMPUs from the cache following receipt of a deregistration request fails, this is passed back to the client as a 500 request, unless another error has already been encountered (the cache delete is attempted even if the request has already failed for some other reason to maximise the chances of tidyup) in which case that preexisting error is reported instead.  If the IMPU delete gets a "Not found" error back from Cassandra, this is treated as though Cassandra had returned OK.
-  There is no attempt to wait for a Cassandra operation performed as part of a Registration Termination Task, as these are driven by the HSS and so a client cannot legitimately be waiting on such a request to complete and cannot be relying on any implicit state change.


steelyvoid:
PutAssociatedPrivateID sync issue is being tracked by https://github.com/Metaswitch/homestead/issues/292

src/handlers.cpp
```diff
@@ -329,19 +329,20 @@ void ImpiTask::send_mar()
 
 void ImpiTask::on_mar_response(Diameter::Message& rsp)
 {
-  Cx::MultimediaAuthAnswer maa(rsp);
+  Cx::MultimediaAuthAnswer *maa = new Cx::MultimediaAuthAnswer(rsp);
```
mirw: Why don't we put the Cx::MultimediaAuthAnswer in the ImpiTask itself?  That would avoid having to remember to delete the Cx::MultimediaAuthAnswer when we're finished.

steelyvoid: I did consider this, but the MAA can't be created until we've got the MAR response, so the MAA object would have be "undefined" until we get to this point.  Is there a C++ construct that lets you do this?

mirw: I thought we allowed MAA's to wrap a NULL pointer, but we don't.  I think we can address my leak concerns, though, by always writing the MultimediaAuthAnswer into _maa (disposing of the local maa variable) and then destroying _maa in this object's destructor.  Does that make sense?

src/handlers.cpp
```diff
@@ -385,6 +398,36 @@ void ImpiTask::on_mar_response(Diameter::Message& rsp)
       break;
   }
 
+  // If not waiting for the assoicated IMPU put to succeed, tidy up now
+  if (!updating_assoc_public_ids)
+  {
+    delete maa;
```
mirw: If we write directly to _maa and destroy it in this task's destructor, we can remove the need for `delete maa` here and elsewhere.

src/handlers.cpp
```diff
@@ -1579,28 +1632,129 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
...
       SAS::report_event(event);
-      send_http_reply(HTTP_SERVER_ERROR);
+      _http_rc = HTTP_SERVER_ERROR;
       break;
   }
-  delete this;
+
+  // If we're not caching the registration data, check for deregistration and
+  // delete the task now
+  if (!caching)
```
mirw: This logic feels a bit confusing.  Isn't the following code "caching" (or rather "uncaching") too?  Can we pull the "cache update" code (either for adding or deleting cache entries) into one place?

src/handlers.cpp
```diff
@@ -1531,40 +1616,8 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
...
-      CassandraStore::Operation* delete_public_id =
-        _cache->create_DeletePublicIDs(public_ids,
-                                       associated_private_ids,
-                                       Cache::generate_timestamp());
-      CassandraStore::Transaction* tsx = new CacheTransaction;
-      _cache->do_async(delete_public_id, tsx);
-    }
-  }
-
+  _http_rc = HTTP_OK;
```
mirw: Why don't we initialize this in the constructor?

src/handlers.cpp
```diff
@@ -1579,28 +1632,129 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
...
+    // If we're not deleting the public IDs, reply and quit now.
+    if (!deleting)
+    {
+      if (_http_rc == HTTP_OK)
+      {
+        send_reply();
+      }
+      else
+      {
+        send_http_reply(_http_rc);
```
mirw: Should we delegate this _http_rc test to send_reply to avoid duplicating it everywhere?

mirw:
Sorry for the delay on this.  A few more comments.  Mainly trying to keep the code tidy.

src/handlers.cpp
steelyvoid: Yup - thanks

src/handlers.cpp
steelyvoid: Yes, that doesn't look right.  If we do a put_in_cache, we'll never do this deregistration stuff.  I'll need to rethink

steelyvoid: Ah - I remember.  We only call put_in_cache at line 1634 if is_deregistration_request(_type) is false, so we don't need to do the code from 1662 to 1694 if we've done a put_in_cache.  Does that make sense?  Do you want me to change the commenting to make this clearer?

src/handlers.cpp
steelyvoid: Will do

src/handlers.cpp
steelyvoid: Happy to do that

src/handlers.cpp
mirw: I was more observing that maybe we should pull the `put_in_cache` at line 1634 to here, so that we have a single block of code responsible for managing the cache, rather than doing some cache management above and then remembering whether we did it and doing the rest of the cache management down here.

steelyvoid: Understood.  Yep - I'm happy to make that change

steelyvoid:
Hi Matt.  Could you cast a final eye over my review updates to check that we're in sync?  Thanks!

src/handlers.cpp
```diff
@@ -162,6 +162,15 @@ void ImpiTask::run()
   }
 }
 
+ImpiTask::~ImpiTask()
+{
+  if (_maa != NULL)
```
mirw: There's no need to check `if (_maa != NULL)` - `delete _maa` implicitly checks that _maa is non-NULL.

src/handlers.cpp
```diff
@@ -1578,29 +1640,120 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
...
+  on_del_impu_benign(op, false);
+}
+
+void ImpuRegDataTask::on_del_impu_failure(CassandraStore::Operation* op, CassandraStore::ResultCode error, std::string& text)
+{
+  // Failed to delete IMPUs. If the error was "Not Found", just pass back the
+  // stored error code.  "Not Found" errors are benign on deletion
+  if (error == CassandraStore::NOT_FOUND)
+  {
+
```
mirw: Spurious blank line.

src/handlers.cpp
```diff
@@ -1578,29 +1640,120 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
       {
         TRC_DEBUG("Getting User-Data from SAA for cache");
         saa.user_data(_xml);
-        put_in_cache();
+        caching = true;
```
mirw: This wasn't quite what I meant.  See https://github.com/Metaswitch/homestead/commit/c1536ce6465a33e4745b5e91ad2b59e40e669478 for what I was suggesting - basically, all the caching logic becomes its own unit, so the processing in this function consists of discrete blocks:

*   Parse the response/extract information.
*   Update the cache.
*   Send the response (or don't if there's a pending cache operation).

This feels more maintainable.

mirw:
3 more comments - only the last is of any substance, and you can just pull in my commit if you're happy with it.

src/handlers.cpp
steelyvoid: Happy with the rearrangement, but I do have one question about your suggested commit.  You have saa.user_data(_xml) being called unconditionally (even if its a dereg or auth failure).  Is this functional change safe?

mirw: Yes, I believe so - what was your concern?

*   The AVP shouldn't be present on deregistration or authentication failure responses.
*   If the AVP is present, we return it on the HTTP response.
*   If the AVP is not present, we omit it from the response (as at present).

PR #301, opened 2016-02-10
[Reviewer: Adam] Move cpp-common UTs to cpp-common

eleanor-merry:
Example of the changes to other repos as part of https://github.com/Metaswitch/cpp-common-test/pull/1
Take out cpp-common files from coverage-not-yet and move UTs to cpp-common


johadalin:
LGTM. 
Do the same elsewhere :)

PR #303, opened 2016-02-24
Reviewer RKD: Retry 3002s at a different node

### LINKED ISSUE #273: If one Homestead out of two loses its connection to the HSS, things fail
I had a setup where:

* I had two Homestead nodes
* Homestead-A had a good connection to the HSS
* Homestead-B had no connection to the HSS (no need to go into why)
* I tried to register a subscriber

On finding it had no connected Diameter peers, Homestead-B returned a 500 error. Sprout didn't retry and logged "Fatal HTTP error encountered". The call failed.

We should have retried to Homestead-A (which would have succeeded). In the cases where it's valid not to retry, we should clearly log that we're not retrying and why.

graemerobertson:
Fixes #273.

As discussed, if Homestead has no HSS connections, freeDiameter returns DIAMETER_UNABLE_TO_DELIVER. We should try a different Homestead node in this case, so return a 503 to Sprout.

I've also moved a SAS log which just looked like it was in the wrong place when I was looking at SAS.


src/handlers.cpp
```diff
@@ -385,6 +385,12 @@ void ImpiTask::on_mar_response(Diameter::Message& rsp)
       }
     }
     break;
+    case DIAMETER_UNABLE_TO_DELIVER:
+      // LCOV_EXCL_START - nothing interesting to UT.
+      TRC_ERROR("Unable to deliver MAR - suggest retrying from a different Homestead node");
```
rkd-msw: I'm not sure this log is useful - I generally find https://github.com/Metaswitch/cpp-common/blob/master/src/diameterstack.cpp#L288 enough to diagnose this issue.

What it might be worth having here is a comment recording why this is the right ting to do (matching our IM discussion - i.e. another Homestead might have a connection to the HSS, or a connection to an SLF which has a connection to the HSS, and returning 503 means Sprout might try that Homestead).

rkd-msw:
One comment, otherwise looks good.

PR #323, opened 2016-03-24
[Reviewer: Rob] Config changes

eleanor-merry:
Homestead changes to match https://github.com/Metaswitch/sprout/pull/1364

### DIFF
--- a/debian/homestead.init.d
+++ b/debian/homestead.init.d
@@ -118,13 +118,16 @@ get_settings()
 
         log_level=2
         num_http_threads=$(($(grep processor /proc/cpuinfo | wc -l) * 50))
-
+      
         # Derive server_name and sprout_http_name from other settings
         if [ -n "$scscf_uri" ]
         then
           server_name=$scscf_uri
+        elif [ -n "$scscf_prefix" ]
+        then
+          server_name="sip:$scscf_prefix.$sprout_hostname:$scscf;transport=TCP"
         else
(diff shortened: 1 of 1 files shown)


debian/homestead.init.d
```diff
@@ -122,8 +122,11 @@ get_settings()
         if [ -n "$scscf_uri" ]
         then
           server_name=$scscf_uri
+        elif [ -n "$scscf_prefix" ]
+        then
+          server_name="sip:$scscf_prefix.$sprout_hostname:$scscf;transport=TCP"
         else
-          server_name="sip:$(python /usr/share/clearwater/bin/bracket_ipv6_address.py $sprout_hostname):$scscf;transport=TCP"
```
rkd-msw: Is this right? Aren't you going to end up with "scscf.10.0.0.3"?

e.g. on a non-AWS AIO node: https://github.com/Metaswitch/clearwater-infrastructure/blob/master/debian/clearwater-auto-config-generic.init.d#L72

rkd-msw:
One comment.

debian/homestead.init.d
rkd-msw: We discussed - this is part of the https://github.com/Metaswitch/sprout/issues/929 weirdness but should 'just work'.

eleanor-merry: It's not that bug - it's just when sprout_hostname isn't a hostname. It should still 'just work'

PR #339, opened 2016-04-26
[Reviewer: Rob] Pass through the S-CSCF server name

### LINKED ISSUE #66: Pass server-name and sprout-http-name through from sprout
We could pass sprout's sip uri and http address to homestead on regdata lookups. The sip uri is used as the server name on SARs and the http address is used to send deregistration information triggered by RTRs. They are currently taken from /etc/clearwater/config and are just sprout@domain.

eleanor-merry:
Allow Sprout to pass the server name through rather than always pulling it from the shared config

Fixes #66 (and is part of the fix for https://github.com/Metaswitch/sprout/issues/1405)

### DIFF
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -53,7 +53,7 @@ const std::string SIP_URI_PRE = "sip:";
 Diameter::Stack* HssCacheTask::_diameter_stack = NULL;
 std::string HssCacheTask::_dest_realm;
 std::string HssCacheTask::_dest_host;
-std::string HssCacheTask::_server_name;
+std::string HssCacheTask::_configured_server_name;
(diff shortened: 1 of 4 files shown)


src/handlers.cpp
```diff
@@ -1483,7 +1484,8 @@ void ImpuRegDataTask::send_server_assignment_request(Cx::ServerAssignmentType ty
                                   _dest_realm,
                                   _impi,
                                   _impu,
-                                  _server_name,
+                                  (_provided_server_name == "" ? _server_name :
```
rkd-msw: Do `_server_name` and `_fallback_server_name` make more sense?

rkd-msw:
LGTM

PR #344, opened 2016-05-05
[Reviewer: Graeme] Move S-CSCF URI to body

eleanor-merry:
Moves the server name to the body

### DIFF
--- a/docs/homestead_api.md
+++ b/docs/homestead_api.md
@@ -80,9 +80,9 @@ The valid values of reqtype are:
 * `dereg-user`, `dereg-timeout`, `dereg-admin` - used to indicate that a deregistration (e.g. a REGISTER with `Expires: 0`, expiry of all bindings, or some other failure) has triggered this request. If a HSS is configured, Clearwater will delete any cached data for the subscriber (putting it in NOT_REGISTERED state) and send a Server-Assignment-Request with an appropriate type (USER_DEREGISTRATION, TIMEOUT_DEREGISTRATION or ADMINISTRATIVE_DEREGISTRATION). If a HSS is not configured, Clearwater will set the user to UNREGISTERED state.
 * `dereg-auth-failure` - used to indicate that a registration of a new binding has failed. This doesn't change any state on Homestead (as a registered user who fails to register a second binding shouldn't be de-registered, and an unregistered or not-registered user who fails to register is already in the right state) - it simply triggers a Server-Assignment-Request with Server-Assignment-Type AUTHENTICATION_FAILURE.
 
-The body also has an optional server_name field, specifying the S-CSCF URI to use on the Server-Assignment-Request.
+These Server-Assignment-Types will specify a User-Name based on the `private_id` query parameter (if provided), or on the PrivateID element in the cached User-Data. It will be omitted if neither of these are available.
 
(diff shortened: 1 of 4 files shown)


docs/homestead_api.md
```diff
@@ -80,7 +80,9 @@ The valid values of reqtype are:
...
 * `dereg-auth-failure` - used to indicate that a registration of a new binding has failed. This doesn't change any state on Homestead (as a registered user who fails to register a second binding shouldn't be de-registered, and an unregistered or not-registered user who fails to register is already in the right state) - it simply triggers a Server-Assignment-Request with Server-Assignment-Type AUTHENTICATION_FAILURE.
 
-These Server-Assignment-Types will specify a User-Name based on the `private_id` query parameter (if provided), or on the PrivateID element in the cached User-Data. It will be omitted if neither of these are available. It will also specify the Server-Name based on the `server_name` query parameter (if provided), or the value of the configured S-CSCF URI (in `/etc/clearwater/shared_config`).
+The body also has an optional server_name field, specifying the S-CSCF URI to use on the Server-Assignment-Request.
```
graemerobertson: Can you put server_name in double quotes for consistency with reqtype above?

docs/homestead_api.md
```diff
@@ -80,7 +80,9 @@ The valid values of reqtype are:
...
+
+These Server-Assignment-Types will specify a User-Name based on the `private_id` query parameter (if provided), or on the PrivateID element in the cached User-Data. It will be omitted if neither of these are available. It will also specify the Server-Name based on the `server_name` field (if provided), or the value of the configured S-CSCF URI (in `/etc/clearwater/shared_config`).
```
graemerobertson: Although these parameters are all in single quotes, so that's nice. Please just make them all consistent.

graemerobertson: Server-Assignment-Types should be Server-Assignment-Requests I think?

src/handlers.cpp
```diff
@@ -1043,14 +1043,34 @@ ImpuRegDataTask::RequestType ImpuRegDataTask::request_type_from_body(std::string
...
+std::string ImpuRegDataTask::server_name_from_body(std::string body)
+{
+  TRC_DEBUG("Determining server name from '%s'", body.c_str());
+
+  rapidjson::Document document;
+  document.Parse<0>(body.c_str());
+
+  if (!document.IsObject() ||
+      !document.HasMember("server_name") ||
+      !document["server_name"].IsString())
```
graemerobertson: This if statement contains a mixture of errors not errors - you could split these up to make debugging easier.

src/handlers.cpp
```diff
@@ -1043,14 +1043,34 @@ ImpuRegDataTask::RequestType ImpuRegDataTask::request_type_from_body(std::string
...
+
 void ImpuRegDataTask::run()
 {
   const std::string prefix = "/impu/";
   std::string path = _req.full_path();
 
   _impu = path.substr(prefix.length(), path.find_first_of("/", prefix.length()) - prefix.length());
   _impi = _req.param("private_id");
-  _provided_server_name = _req.param("server_name");
+  _provided_server_name = server_name_from_body(_req.get_rx_body());
```
graemerobertson: It's weird to put this above the log on the line below and not include it in the log below. In fact, I imagine the debug logs for this section of code are an absolute mess.

server_name_from_body logs the HTTP body, then we have the log below with the private ID and public ID, and then on PUTs request_type_from_body logs the HTTP body again.

I think we should either log the HTTP body once or just log what we've gleaned from the HTTP body.

graemerobertson:
Reviewed

PR #348, opened 2016-05-12
[Reviewer: Matt] Tweak the Cassandra is_registered column so we heal better from netsplits

### LINKED ISSUE #345: GR (split-brain) callee's state is set to "unregistered" in HSS after cache failure
#### Symptoms

On a GR system where communications has been cut between sites, if a UE from one site calls a UE in the other site, the call fails appropriately but the callee's state in the HSS gets set to "unregistered" by homestead.

#### Impact

Once communications between the sites is restored, the callee can't receive any calls until it re-registers.

#### Release and environment 

release-94

#### Steps to reproduce
1, Start with two unregistered UEs (softphones)
2. Cut communications between geo sites
3. Register both UEs
4. Try calling for one site to the other
5. Call fails with a 480. However, if I look at the state of the “callee” in the HSS, has been changed from registered to unregistered (not to be confused with NOT-REGISTERED). 
6. Restore communications between the two sites.
7. Any attempt to call the “callee” (from either site) results in a 480 until the callee re-registers. This seems bad especially if the UEs re-registration time is 30 minutes, since this would mean that the user would be blissfully ignorant that he couldn’t receive a for that long. Note: The “callee” can make calls during this time.

A SAS trace for the call in step#5 is in the diags area.

If you look in that SAS trace, you’ll see that it’s homestead deciding to change the registration state to “unregistered”:
p1
p2

rkd-msw:
I think this fixes #345.

I should be able to repro the problem (and then test that it's fixed) by creating 2 GR sites through Chef and following the repro instructions in that issue.

### DIFF
--- a/include/handlers.h
+++ b/include/handlers.h
@@ -504,6 +504,7 @@ protected:
   std::string _type_param;
   RequestType _type;
   std::string _xml;
+  RegistrationState _original_state;
   RegistrationState _new_state;
   ChargingAddresses _charging_addrs;
   long _http_rc;
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -1622,7 +1622,10 @@ void ImpuRegDataTask::put_in_cache()
       // Using a timestamp in the past for the UNREGISTERED case means that,
       // when we do the last-write-wins conflict resolution, we'll always make
       // the REGISTERED state win.
-      timestamp -= (_cfg->hss_reregistration_time * 2);
+      //
+      // Multiply by 1,000,000 - timestamps are in microseconds
+      int64_t diff = (int64_t)_cfg->hss_reregistration_time * 2 * 1000000;
+      timestamp = timestamp - diff;
     }
 
     Cache::PutRegData* put_reg_data = _cache->create_PutRegData(public_ids,
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -1607,33 +1607,23 @@ void ImpuRegDataTask::put_in_cache()
     event.add_var_param(_charging_addrs.log_string());
     SAS::report_event(event);
 
-    int64_t timestamp = Cache::generate_timestamp();
-
-    if ((_original_state == RegistrationState::NOT_REGISTERED) &&
-        (_new_state == RegistrationState::UNREGISTERED))
-    {
-      // Force the timestamp to be at least two re-registration intervals in
-      // the past. This handles the case where:
-      // * there is a netsplit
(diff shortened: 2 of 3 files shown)


mirw:
Remind me in what scenario we go to state UNREGISTERED?  I think we can go from REGISTERED to UNREGISTERED due to an explicit un-REGISTER.  In that case, I think we want the UNREGISTERED state to win, not lose.  That would suggest we need to look at what the old state was, and only set an earlier timestamp if we're transitioning from NOT_REGISTERED to UNREGISTERED.  Thoughts?

(Apart from that, code looks sensible.)

rkd-msw:
I think an explicit un-REGISTER moves us to NOT_REGISTERED, but the change you've suggested is sensible and protects us if we do want to send one of the _STORE_SERVER_NAME Server-Assignment-Types in future (which would just move us to UNREGISTERED), so I'll make it.

rkd-msw:
I've tested this:

* using UNREGISTERED timestamps in the past fixes the repro scenario in #345.
* however, when a subscriber explicitly de-registers, their tombstone marker has a current timestamp, so that wins out over any UNREGISTERED records we write for a while (~1 hour by default), meaning that we don't cache the IFCs for that user and re-query the HSS for every call. This happens even in non-split brain scenarios.
* we can't just move the tombstone record into the past, because it needs to win out over explicit registrations.

Also, the user is still recorded as UNREGISTERED in the HSS, because that's the last Server-Assignment-Request the HSS saw - we don't do any active fixup towards the HSS when our Cassandra cluster rejoins.

Given that, I'm nervous about merging this - it fixes an edge/error case, but not perfectly (wrong state in the HSS) and causes regressions in the mainline case.

mirw:
Agreed.

As discussed, should we separate assignment and registration into separate columns?

rkd-msw:
I've implemented the approach from https://github.com/Metaswitch/homestead/issues/345#issuecomment-223551402 which works:

* I've verified that the is_registered column is now `null`, not `False`, for unregistered subscribers but is still `True` for registered subscribers
* the repro scenario in #345 is fixed - after I rejoin the sites, both subscribers are marked as registered and calls between them work
* the issue with my previous fix no longer happens - iFCs for an unregistered subscriber are cached, and I've checked with tshark that we don't send a SAR on subsequent calls to them

As none of the lines of code you originally reviewed have survived, it's worth you re-reviewing before I merge.

mirw:
Looks good.  No comments.

Issue #324, opened 2016-04-01
make full_test fails with valgrind error


dtheobald:
#### Symptoms

When running ```make full_test``` I get the following error:
```
Error: some memory errors have been detected
Leak_DefinitelyLost
Leak_DefinitelyLost
See ../build/homestead_test/valgrind_output.xml for further details.
make[1]: *** [valgrind_check_homestead_test] Error 2
make[1]: Leaving directory `/code/CCv9/cc/homestead/src'
make: *** [homestead_full_test] Error 2
```
[valgrind_output.txt](https://github.com/Metaswitch/homestead/files/199740/valgrind_output.txt)

#### Release and environment 

release-94

#### Steps to reproduce

1) make full_test

JimBunch:
I need more information about how this issue affects call processing or system operation before we can determine if it is critical or not.

mirw:
I would suggest it's not critical (apart from that it may mask regressions in future).  The memory leak is part of gnutls initialization, and appears to be that gnutls is not tidying up correctly.

However, we don't see this on our continuous integration builds - you can see the output from these at http://repo.cw-ngv.com:8080/job/homestead/ws/build/homestead_test/valgrind_output.xml, so I think we need to understand what the difference is between your build environment and our continuous integration builds.

I've checked that the continuous integration build server is using the same version of the libgnutls.so library as on your build machine (visible in the log).

Is the code that you're compiling identical to the dev branch?  (It looks like it might be, but please can you confirm?)

Was the code cleaned before the `make full_test`?  (It should have rebuilt everything that had changed anyway, but it's worth checking in case the cause of the difference.)

Is the command-line output of the test run exactly the same as that from http://repo.cw-ngv.com:8080/job/homestead/103/console (search for "valgrind" to find the valgrind test output).  I'm just wondering if one of the tests ends in a different way on your build system, and so doesn't terminate gnutls cleanly.

dtheobald:
The code is the same as release-94 plus the changes from the release-94-fixes branch.

Yes, the ```make full_test``` was done on clean directory and I've attached the valgrind output.
[valgrind_output.txt](https://github.com/Metaswitch/homestead/files/286363/valgrind_output.txt)

bossmc:
This is a known issue (@steelyvoid - I think you've added a suppression for this in various repos?), the underlying issue is a race condition between libdl and pthreads (more accurately it's UB, so some builds will consistently hit the issue, while others won't).

[...]

PR #351, opened 2016-05-27
[Reviewer: Steve] Apply valgrind suppression

mirw:
Steve,

Please can you review this fix to #324 (to suppress valgrind errors in gnutls when used from freeDiameter)?

Thanks,

Matt


steelyvoid:
Looks good.  I'm surprised you needed the homestead_test_VALGRIND_ARGS := --suppressions=ut/homestead_test.supp line (since ut/homestead_test.supp already existed), but if the bottom line is that the gnutls leaks go away, I'm happy

mirw:
Thanks!  I'm not sure the ut/homestead_test.supp file did exist before.  Either way, it seems to be required.

eleanor-merry:
FYI, Homestead's suppression file was moved to the new cpp-common-test repo (so older versions of Homestead still have it)

PR #354, opened 2016-06-17
[Reviewer AJH]: SAS trace HSS returncodes

steelyvoid:
Added tracing of Diameter returncodes to SAS.

Tested using an HSS that was rejecting deregistrations with a 5003 DIAMETER_AUTHORIZATION_REJECTED error

### DIFF
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -689,20 +689,20 @@ void ImpiRegistrationStatusTask::on_uar_response(Diameter::Message& rsp)
            (experimental_result_code == DIAMETER_ERROR_IDENTITIES_DONT_MATCH))
   {
     TRC_INFO("User unknown or public/private ID conflict - reject");
-    sas_log_hss_failure(experimental_result_code);
+    sas_log_hss_failure(result_code, experimental_result_code);
     send_http_reply(HTTP_NOT_FOUND);
   }
   else if ((result_code == DIAMETER_AUTHORIZATION_REJECTED) ||
            (experimental_result_code == DIAMETER_ERROR_ROAMING_NOT_ALLOWED))
   {
     TRC_INFO("Authorization rejected due to roaming not allowed - reject");
-    sas_log_hss_failure(result_code ? result_code : experimental_result_code);
+    sas_log_hss_failure(result_code, experimental_result_code);
     send_http_reply(HTTP_FORBIDDEN);
   }
   else if (result_code == DIAMETER_TOO_BUSY)
   {
     TRC_INFO("HSS busy - reject");
-    sas_log_hss_failure(result_code);
+    sas_log_hss_failure(result_code, experimental_result_code);
     send_http_reply(HTTP_GATEWAY_TIMEOUT);
   }
   else if (result_code == DIAMETER_UNABLE_TO_DELIVER)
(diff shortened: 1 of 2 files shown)


AlexHockey:
Thanks @steelyvoid. TBH, I don;t see why this PR is needed or is the right thing to do. I have a couple of concerns with it:

* You include the result code in a high level log, but you present it basically as a protocol value (e.g. `DIAMETER_SUCCESS`) rather than as a description of what might have gone wrong. If the user is interested in protocol details why wouldn't they just look at the protocol flows?
* `sas_log_hss_failure` can be called either with a result code or an experimental result code. But https://github.com/Metaswitch/clearwater-sas-resources/pull/100 assumes that all values are plain result codes, so it will crack experimental result codes out incorrectly. 

Thoughts?

steelyvoid:
In the instance I was looking at that prompted this change, there were actually several Diameter flows for the subscriber in the SAS trace and it wasn't obvious from the looking at the trace which one corresponded to the HSS_FAIL log.  Including the return code in the log seemed to be a reasonably cheap way to simplify the searching logic here.  I was also concerned that, in the event that the underlying problem is that the HSS is uncontactable, you won't get any protocol trace, and seeing the returncode we have in hand will allow a reader to get to the bottom of the problem quickly.

On the sas_log_hss_failure point, I had missed that this was being called with different versions of the result code. Would this worry be addressed if I were to change the sas_log_hss_failure routines to take and trace both result and experimental result (and only attempt to map the main result code to a string)?

AlexHockey:
I guess I don't really see why you need to know what flow triggered the HSS failed log. That is a high level log meant to steer users to look at the HSS interactions. Isn't it then obvious from the flows what operation failed?

Note that if you don't get any protocol trace you don't get any result code to log either!

steelyvoid:
For me its more a question of useability.  If we have a failure log, and the reason for the failure is in our hand, why make the reader hunt around in the protocol trace looking for the relevant field?  If you can minimise the chance of a SAS user making an unnecessary support call because they haven't spotted the error field in the protocol trace, surely thats a good thing?

On the second point above, is it always going to be the case that we'll only ever make that HSS_FAIL log if we have successfully send out traffic and got a failure response?  Are there really no errors that might be reported by the freeDiameter stack on the client side?

steelyvoid:
Alex, we agreed that the ideal fix here would be for the result codes to appear in the ladder diagram and perhaps highlighted in the message itself to give the same experience as other protocols in SAS, but, as this is a significant fix to SAS decoders, I should raise an issue for this and, for the short term, I should go ahead and put the error codes in the SAS messages themselves with some text to make it clear that this MIGHT be the reason for the HSS failure.

How does the following detail text look to you

"The following result codes reported by the Diameter stack may help to identify the root cause of the failure:
Result: ....
Experimental result: ..."

AlexHockey:
That looks good, thanks Steve. There might be some finessing you can do in the UI (to make the Result/Experimental Result only appear if they are set in SAS event) but I'll leave that up to you if you want to bother with that!

AlexHockey:
Passing back to you while you implement the new approach.

steelyvoid:
Hi Alex.  Can you double check that this is what you were expecting?

Do you want me to update SAS timestamps as part of this fix?

src/handlers.cpp
```diff
@@ -1700,19 +1706,12 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
...
       break;
       // LCOV_EXCL_STOP
-    case 5001:
-    {
-      TRC_INFO("Server-Assignment answer with result code %d - reject", result_code);
-      SAS::Event event(this->trail(), SASEvent::REG_DATA_HSS_FAIL, 0);
-      SAS::report_event(event);
-      _http_rc = HTTP_NOT_FOUND;
-    }
-    break;
```
AlexHockey: I notice you've removed this branch. The code as it stands is catching a 5001 which is  `DIAMETER_AVP_UNSUPPORTED`. However the 3GPP experimental result code 5001 is `DIAMETER_ERROR_USER_UNKNOWN` which is what I think this code is trying to handle. This makes me think that a SAR can have an experimental result code which we're not logging. It also suggests the code is bugged because its handling the wrong result code field. Please can you:

* Check whether SARs can have an experimental result code. 
* If they can, log it to SAS and display it. 
* Raise an issue covering the fact that this code looks wrong.

AlexHockey:
Looks good, one comment inline.

PR #357, opened 2016-07-04
[Reviewer: MIRW] Use common daemon setup/logging

richardwhiuk:
This adds standard error / standard output logging to homestead.


mirw:
Looks good.

PR #361, opened 2016-07-15
[Reviewer: AJH] Allow an external Cassandra to be configured

richardwhiuk:
Alex,

This is the change required to allow homestead to use external storage. We need to pass in the documented configuration option `cassandra_hostname` into Homestead, which already has the code to use it if passed in.

We also need to fix the keyspace creation files to work with an external cassandra by skipping the filesystem check as to whether the keyspace already exists, and pointing cqlsh at the external Cassandra cluster.


homestead.root/usr/share/clearwater/cassandra-schemas/homestead_cache.sh
```diff
@@ -1,11 +1,17 @@
...
 . /etc/clearwater/config
 
 . /usr/share/clearwater/cassandra_schema_utils.sh
 
 quit_if_no_cassandra
 
-if [[ ! -e /var/lib/cassandra/data/homestead_cache ]];
+CQLSH="/usr/share/clearwater/bin/run-in-signaling-namespace cqlsh $cassandra_hostname"
+
+if [[ ! -e /var/lib/cassandra/data/homestead_provisiong ]] || \
```
AlexHockey: Homestead shouldn't be checking the provisioning database since it doesn't use it. But it *definitely* shouldn't be checking the "provisiong" database ;-)

homestead.root/usr/share/clearwater/cassandra-schemas/homestead_cache.sh
```diff
@@ -1,11 +1,17 @@
...
+   [[ cassandra_hostname -ne "127.0.0.1" ]];
```
AlexHockey: Missing `$`?

AlexHockey: `-ne` is for integer comparison.

AlexHockey:
Hi Richard, aside from the typos, I'm not keen on the way that if cassandra is on a different box, the script waits for a while for the local cassandra to come up (which is basically irrelevant), then talks to the remove cassandra. Can you refactor this so:

* We only wait for the local cassandra if the hostname is local.
* We use cqlsh to check if the schema has been programmed (could either do this by querying the system keyspace or by trying to use the homestead_cache keyspace and seeing if there are any errors.

bossmc:
We've considered in the past using migrations (e.g. https://github.com/comeara/pillar) to better manage schema definitions (which might also help with upgrade/downgrade on future too).

AlexHockey:
Yeah but for now we should do something pragmatic.

richardwhiuk:
We don't wait for the local cassandra to come up - we wait for the remote cassandra to come up - /usr/share/clearwater/bin/poll_cassandra.sh uses $cassandra_hostname, which is the remote cassandra cluster. Most of the code was happy to use a remote cassandra - it seems to have been half implemented.

Arguably we should do something about the keyspace already existing - currently it just fails with an error when we attempt to create it when it already exists, but that's kind of fine.

AlexHockey:
Cool, in that case we can run the wait code unconditionally.

> Arguably we should do something about the keyspace already existing - currently it just fails with an error when we attempt to create it when it already exists, but that's kind of fine.

What happens when you get this failure. Does it appear anywhere the user or support might see it (e.g. in logs)? Doe sit cause the script to bail out early. If yes, I think we should fix it. 

Additional comments inline.

homestead.root/usr/share/clearwater/cassandra-schemas/homestead_cache.sh
```diff
@@ -1,11 +1,17 @@
...
 . /etc/clearwater/config
 
 . /usr/share/clearwater/cassandra_schema_utils.sh
 
 quit_if_no_cassandra
 
-if [[ ! -e /var/lib/cassandra/data/homestead_cache ]];
+CQLSH="/usr/share/clearwater/bin/run-in-signaling-namespace cqlsh $cassandra_hostname"
+
+if [[ ! -e /var/lib/cassandra/data/homestead_case ]] || \
```
AlexHockey: Another typo.

debian/control
```diff
@@ -24,7 +24,7 @@ Description: Debugging symbols for homestead-libs
 
 Package: homestead
 Architecture: any
-Depends: clearwater-infrastructure, clearwater-tcp-scalability, clearwater-log-cleanup, homestead-libs, libboost-regex1.54.0, libzmq3, libcurl3-gnutls, gnutls-bin, clearwater-socket-factory, libboost-filesystem1.54.0, libsnmp30 (>= 5.7.2~dfsg-clearwater4), clearwater-cassandra, clearwater-monit
+Depends: clearwater-infrastructure, clearwater-tcp-scalability, clearwater-log-cleanup, homestead-libs, libboost-regex1.54.0, libzmq3, libcurl3-gnutls, gnutls-bin, clearwater-socket-factory, libboost-filesystem1.54.0, libsnmp30 (>= 5.7.2~dfsg-clearwater4), clearwater-monit
```
AlexHockey: You removed the dependency on clearwater-cassandra. Isn't this going to break standard Project Clearwater users? Hw does cqlsh get installed?

AlexHockey:
Back to you to answer my question on the impact of errors.

Given there were a few potential regressions in this PR, I think you should test that:

* A traditional clearwater fresh install still works without errors. 
* Upgrading an existing deployment to one with this new code works with no errors.

richardwhiuk:
Regarding impact of errors, on upgrading you see the following text:

```
<stdin>:2:Bad Request: Cannot add existing keyspace "homestead_cache"
<stdin>:9:Bad Request: Cannot add already existing column family "impi" to keyspace "homestead_cache"
<stdin>:14:Bad Request: Cannot add already existing column family "impu" to keyspace "homestead_cache"
<stdin>:3:Bad Request: Cannot add already existing column family "impi_mapping" to keyspace "homestead_cache"`
```

and then the upgrade proceeds as normal. I'll double check a standard deployment shortly.

PR #363, opened 2016-07-15
[Reviewer AJH] Allow SAS traffic to be optionally routed via signaling namespace

steelyvoid:
Adds option to route SAS traffic via the signaling network


AlexHockey:
:+1:

PR #365, opened 2016-08-02
[Reviewer: AJL] Split storage debian packaging for Homestead

sathiyan-sivathas:
(3/6)


debian/homestead.postinst
```diff
@@ -89,13 +89,6 @@ case "$1" in
 
         service clearwater-infrastructure restart
 
-        # Stop the cluster manager, so that it is restarted by Monit and picks
-        # up the new Cassandra scaling plugin. We check whether the process is 
-        # installed to avoid warning logs when etcd isn't being used
-        if [ -x "/etc/init.d/clearwater-cluster-manager" ]; then
-          service clearwater-cluster-manager stop || /bin/true
-        fi  
-
```
johadalin: Did you discuss this with Rob or somebody else? 
I don't know enough to say this is still needed, but given that these postinst files are going to be in place on -node nodes as well as nodes using the smaller packages, we may well want to keep this in.

johadalin:
One comment, otherwise i think fine :+1:

debian/homestead.postinst
sathiyan-sivathas: I did discuss with Rob, and we decided to remove this for a reason that wasn't to do with split-storage.  If you look here https://github.com/Metaswitch/clearwater-cassandra/blob/master/debian/clearwater-cassandra.postinst#L72 you'll see that we've already stopped the cluster manager in clearwater-cassandra so we don't need to do it again here.  I suspect that this code is from before cassandra was in its own repo and we forgot to remove it.

rkd-msw: Commit https://github.com/Metaswitch/homestead/commit/be89ba195b8ade458157bc9e58edd7a6c6094648 moved that plugin out but didn't delete this code.

Issue #360, opened 2016-07-11
Homestead can't be configured to send a Server-Assignment-Request on every re-registration


rkd-msw:
#### Symptoms

The IMS specs suggest that whenever a subscriber re-registers, the S-CSCF should send a Server-Assignment-Request with type RE_REGISTRATION to the HSS. Homestead doesn't do this on every re-register - it only does so after hss_reregistration_time seconds have passed, to avoid spamming the HSS.

This is reasonable, but there's no way to configure Homestead to have the spec-compliant behaviour of always sending a SAR - setting a low hss_reregistration_time value doesn't work, as Cassandra TTLs are keyed off this, and setting it too low causes this data to expire before the registration ends.

#### Impact

If the HSS and S-CSCF state get out of sync, re-registering may not fix this, as it might not trigger a SAR.

#### Release and environment 

Current release (Horsea/101).

#### Steps to reproduce

Register a subscriber. Immediately re-register. Observe that no SAR is sent.

rkd-msw:
@steelyvoid

richardwhiuk:
This is fixed in the next release - you can achieve spec compliant behaviour by setting hss_reregistration_time to 0, and homestead will preserve the record for reg_max_expires + 10. 

@steelyvoid FYI

PR #366, opened 2016-08-03
[Reviewer: RKD] Ensure records are preserved until binding expires

richardwhiuk:
Fixes issue #360 as hss_reregistration_timeout can be configured to
0 to force all re-registrations to result in a SAR/SAA.

### DIFF
--- a/debian/homestead.init.d
+++ b/debian/homestead.init.d
@@ -190,6 +190,7 @@ get_daemon_args()
                      --server-name=\"$server_name\"
                      --impu-cache-ttl=$impu_cache_ttl
                      --hss-reregistration-time=$hss_reregistration_time
+                     --reg-max-expires=$reg_max_expires
                      --sprout-http-name=$sprout_http_name
                      $scheme_args
                      $diameter_timeout_ms_arg
--- a/debian/homestead.init.d
+++ b/debian/homestead.init.d
@@ -100,23 +100,11 @@ get_settings()
         scscf=5054
         impu_cache_ttl=0
         max_peers=2
+        hss_reregistration_time=1800
         reg_max_expires=300
         log_level=2
         num_http_threads=$(($(grep processor /proc/cpuinfo | wc -l) * 50))
         . /etc/clearwater/config
-
-        if [ -z $hss_reregistration_time ]
-        then
-          # Default hss_reregistration_time to max(1800, $reg_max_expires/2), so
-          # that homestead only discards expired registration data when sprout
-          # does (at the earliest).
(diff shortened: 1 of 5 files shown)


debian/homestead.init.d
```diff
@@ -190,6 +190,7 @@ get_daemon_args()
                      --server-name=\"$server_name\"
                      --impu-cache-ttl=$impu_cache_ttl
                      --hss-reregistration-time=$hss_reregistration_time
+                     --reg-max-expires=$reg_max_expires
```
rkd-msw: You should probably also get rid of the code at https://github.com/Metaswitch/homestead/blob/52ec9afcf653fbc7b890a954b9858ed23af27976/debian/homestead.init.d#L108 which protected against hss_reregistration_time being below reg_max_expires - I don't think it's helpful now. (Did you test live?)

src/handlers.cpp
```diff
@@ -1192,6 +1192,11 @@ void ImpuRegDataTask::on_get_reg_data_success(CassandraStore::Operation* op)
   // Split the processing depending on whether an HSS is configured.
   if (_cfg->hss_configured)
   {
+    // We need the record to last twice the HSS Re-registration
+    // time, or the max expiry of the registration, which ever one
+    // is longer.
+    int record_ttl = std::max(2 * _cfg->hss_reregistration_time, _cfg->reg_max_expires + 10);
```
rkd-msw: Should you calculate this once in main.cpp and put it into `_cfg->record_ttl`, instead of doing the calculation in a couple of separate functions?

src/handlers.cpp
```diff
@@ -1192,6 +1192,11 @@ void ImpuRegDataTask::on_get_reg_data_success(CassandraStore::Operation* op)
   // Split the processing depending on whether an HSS is configured.
   if (_cfg->hss_configured)
   {
+    // We need the record to last twice the HSS Re-registration
+    // time, or the max expiry of the registration, which ever one
```
rkd-msw: *whichever?

src/handlers.cpp
rkd-msw: Also, comment on the +10?

rkd-msw:
Looks good, couple of comments.

debian/homestead.init.d
richardwhiuk: Yes, I did test live - that code only has any effect if the hss_reregistration time isn't set, and I explicitly set it to 0. I've removed it anyway, as it's no longer required, and unhelpful.

src/handlers.cpp
richardwhiuk: Done

PR #372, opened 2016-09-21
[Reviewer: Ellie] Add Homestead-side changes for AKAv2 support

rkd-msw:
I'll test live before merging.


include/handlers.h
```diff
@@ -298,20 +299,23 @@ class ImpiTask : public HssCacheTask
            int _impu_cache_ttl = 0,
            std::string _scheme_unknown = "Unknown",
            std::string _scheme_digest = "SIP Digest",
-           std::string _scheme_aka = "Digest-AKAv1-MD5",
```
eleanor-merry: There's been a default value set for the schemes in three different places now

debian/homestead.init.d
```diff
@@ -152,9 +158,7 @@ get_daemon_args()
           diameter_timeout_ms=$(( ($target_latency_us + 499)/500 ))
         fi
 
-        [ "$hss_mar_lowercase_unknown" != "Y" ] || scheme_args="--scheme-unknown=unknown"
```
eleanor-merry: Are these config options not still useful?

eleanor-merry:
LGTM - comments inline

debian/homestead.init.d
rkd-msw: I've basically replaced them with more flexible ones - if you wanted this option, you can now say `hss_mar_scheme_unknown="unknown"` instead.

include/handlers.h
rkd-msw: Yes - I thought this would be difficult to fix up, but on reflection I can probably make it be set in only two places, and maybe just in one.

PR #373, opened 2016-09-22
[Reviewer: Alex] Enhance Cassandra connections

sebrexmetaswitch:
Changes to homestead to use the new Cassandra connections

### DIFF
--- a/src/ut/cache_test.cpp
+++ b/src/ut/cache_test.cpp
@@ -42,11 +42,15 @@
 #include "test_interposer.hpp"
 #include "fakelogger.h"
 
+#include "mockhttpresolver.h"
+#include "mock_cassandra_connection_pool.h"
 #include "mock_cassandra_store.h"
 #include "mockcommunicationmonitor.h"
 #include "cass_test_utils.h"
 
 #include <cache.h>
+
+typedef MockHttpResolver MockCassandraResolver;
 
 using ::testing::PrintToString;
 using ::testing::Return;
@@ -82,14 +86,17 @@ const ChargingAddresses ECFS_CHARGING_ADDRS(CCF, ECFS);
 
 // The class under test.
 //
-// We don't test the Cache class directly as we need to override the get_client
-// and release_client methods to use MockCassandraClient.  However all other
-// methods are the real ones from Cache.
+// We don't test the Cache class directly as we need to use a
+// MockCassandraConnectionPool that we can use to return MockCassandraClients.
+// However all other methods are the real ones from Cache.
 class TestCache : public Cache
 {
 public:
-  MOCK_METHOD0(get_client, CassandraStore::Client*());
-  MOCK_METHOD0(release_client, void());
+  void set_conn_pool(CassandraConnectionPool* pool)
+  {
+    delete _conn_pool;
+    _conn_pool = pool;
+  }
 };
 
 //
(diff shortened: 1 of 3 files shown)


src/ut/cache_test.cpp
```diff
@@ -42,12 +42,16 @@
...
 
+#include "mockhttpresolver.h"
+#include "fake_cassandra_connection_pool.h"
 #include "mock_cassandra_store.h"
 #include "mockcommunicationmonitor.h"
 #include "cass_test_utils.h"
 
 #include <cache.h>
 
+typedef MockHttpResolver MockCassandraResolver;
```
AlexHockey: Ewww. I think the mocks should mirror what we're doing with the real classes (e.g. if we have an ARecordResolver and two typedefs, we should also have a MockARecordResolver and two additional typedefs).

src/ut/cache_test.cpp
```diff
@@ -115,12 +137,27 @@ class CacheInitializationTest : public ::testing::Test
...
   MockCassandraClient _client;
+  FakeCassandraConnectionPool* _pool = new FakeCassandraConnectionPool();
+  MockCassandraResolver _resolver;
+  SimpleAddrIterator* _iter;
   AlarmManager* _am = new AlarmManager();
   NiceMock<MockCommunicationMonitor>* _cm = new NiceMock<MockCommunicationMonitor>(_am);
+
+  // Some dummy targets for our resolver
+  std::vector<AddrInfo> _targets;
+  AddrInfo create_target(std::string address)
```
AlexHockey: Need a blank line before the method definition.

src/ut/cache_test.cpp
```diff
@@ -133,9 +170,9 @@ class CacheRequestTest : public CacheInitializationTest
   {
     sem_init(&_sem, 0, 0);
 
-    // By default the cache just serves up the mock client each time.
-    EXPECT_CALL(_cache, get_client()).WillRepeatedly(Return(&_client));
-    EXPECT_CALL(_cache, release_client()).WillRepeatedly(Return());
+    // In some tests we check for success() explicitly, but  this removes the
+    // warning for those where we don't
+    EXPECT_CALL(_resolver, success(_)).Times(testing::AnyNumber());
```
AlexHockey: What about if we have a test where we don't expect success to be called, but it is. I don't think we'll spot such a bug. I guess we could explicitly clear the mock though.

src/ut/cache_test.cpp
```diff
@@ -218,11 +255,29 @@ TEST_F(CacheInitializationTest, Mainline)
...
 {
   apache::thrift::transport::TTransportException te;
-  EXPECT_CALL(_cache, get_client()).Times(1).WillOnce(Throw(te));
-  EXPECT_CALL(_cache, release_client()).Times(0);
+  EXPECT_CALL(_client, connect()).Times(2).WillOnce(Throw(te)).WillRepeatedly(Return());
+
+  {
+    testing::InSequence s;
+
+    EXPECT_CALL(_resolver, blacklist(_targets[0])).Times(1);
```
AlexHockey: I think `Times(1)` is implied if you don;t specify a number of times.

src/ut/cache_test.cpp
```diff
@@ -232,19 +287,41 @@ TEST_F(CacheInitializationTest, TransportException)
 TEST_F(CacheInitializationTest, NotFoundException)
 {
   cass::NotFoundException nfe;
-  EXPECT_CALL(_cache, get_client()).Times(1).WillOnce(Throw(nfe));
-  EXPECT_CALL(_cache, release_client()).Times(0);
+  EXPECT_CALL(_client, set_keyspace(_)).Times(1).WillOnce(Throw(nfe));
+  EXPECT_CALL(_resolver, success(_targets[0])).Times(1);
```
AlexHockey: Worth explaining in a comment why we call success even though `connection_test` returns `NOT_FOUND`.

src/ut/cache_test.cpp
AlexHockey: But you basically have an expect call in every testcase anyway, so do you actually need this?

src/ut/cache_test.cpp
```diff
@@ -452,38 +535,84 @@ MATCHER_P(OperationHasResult, expected_rc, "")
...
   EXPECT_CALL(_client, batch_mutate(_, _)).Times(2).WillRepeatedly(Throw(te));
 
   EXPECT_CALL(*trx, on_failure(OperationHasResult(CassandraStore::CONNECTION_ERROR)));
   EXPECT_CALL(*_cm, inform_failure(_));
   execute_trx((CassandraStore::Operation*)put_reg_data, trx);
 }
 
-TEST_F(CacheRequestTest, PutTransportGetClientEx)
+
+TEST_F(CacheRequestTest, PutTransportConnectEx)
```
AlexHockey: This test throws two exceptions in different places. The testcase name could do with clarifying.

src/ut/cache_test.cpp
```diff
@@ -452,38 +535,84 @@ MATCHER_P(OperationHasResult, expected_rc, "")
...
+  EXPECT_CALL(_resolver, blacklist(_targets[0])).Times(1);
+  EXPECT_CALL(_resolver, blacklist(_targets[1])).Times(1);
 
   EXPECT_CALL(*trx, on_failure(OperationHasResult(CassandraStore::CONNECTION_ERROR)));
   EXPECT_CALL(*_cm, inform_failure(_));
   execute_trx((CassandraStore::Operation*)put_reg_data, trx);
 }
 
+
+TEST_F(CacheRequestTest, PutTransportUnknownException)
```
AlexHockey: Clarify that this test is throwing two exceptions (e.g. PutTransportTHenUnknowException).

PR #375, opened 2016-10-07
[Reviewer: GDR] Add cached data management API. 

AlexHockey:
Hi Graeme, please can you review this change to add an HTTP management API to homestead to allow users to query a subscriber's IMS subscription, and to list all IMPUs homestead is storing data for.

I've tested this in UT, and also live using homestead prov:

* Try to get the subscription for a subscriber who doesn't exist. 
* Get the subscription of a subscriber that does exist. 
* List subscribers when none are configured. 
* List subscribers when some are configured. 

Graeme is doing a fuller live test.

### DIFF
--- a/homestead.root/usr/share/clearwater/infrastructure/scripts/create-homestead-nginx-config
+++ b/homestead.root/usr/share/clearwater/infrastructure/scripts/create-homestead-nginx-config
@@ -1,0 +1,83 @@
+#!/bin/bash
+
+# @file create-homestead-nginx-config
+#
+# Project Clearwater - IMS in the Cloud
+# Copyright (C) 2016 Metaswitch Networks Ltd
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or (at your
+# option) any later version, along with the "Special Exception" for use of
+# the program along with SSL, set forth below. This program is distributed
+# in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+# without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details. You should have received a copy of the GNU General Public
+# License along with this program.  If not, see
+# <http://www.gnu.org/licenses/>.
+#
+# The author can be reached by email at clearwater@metaswitch.com or by
+# post at Metaswitch Networks Ltd, 100 Church St, Enfield EN2 6BQ, UK
+#
+# Special Exception
+# Metaswitch Networks Ltd  grants you permission to copy, modify,
+# propagate, and distribute a work formed by combining OpenSSL with The
+# Software, or a work derivative of such a combination, even if such
+# copying, modification, propagation, or distribution would otherwise
+# violate the terms of the GPL. You must comply with the GPL in all
+# respects for all of the code used other than OpenSSL.
+# "OpenSSL" means OpenSSL toolkit software distributed by the OpenSSL
+# Project and licensed under the OpenSSL Licenses, or a work based on such
+# software and licensed under the OpenSSL Licenses.
+# "OpenSSL Licenses" means the OpenSSL License and Original SSLeay License
+# under which the OpenSSL Project distributes the OpenSSL toolkit software,
+# as those licenses appear in the file LICENSE-OPENSSL.
+
+# This file creates an nginx config file for homestead.
+
+. /etc/clearwater/config
+
+if [ -n "$hs_hostname" ]
+then
+  site_file=/etc/nginx/sites-available/homestead
+  enabled_file=/etc/nginx/sites-enabled/homestead
+  temp_file=$(mktemp homestead.nginx.XXXXXXXX)
+
+  cat > $temp_file << EOF
+upstream http_homestead {
+        server unix:/tmp/homestead-http-mgmt-socket;
+
+        # The minimum number of idle connections to keep alive to the upstream.
+        keepalive 16;
+}
+
+server {
+        listen       [::]:8886 ipv6only=off;
+        server_name  $hs_hostname;
+
+        location / {
+                proxy_pass http://http_homestead;
+                proxy_http_version 1.1;
+
+                # The client may have instructed the server to close the
+                # connection - do not forward this upstream.
+                proxy_set_header Connection "";
+        }
+}
+EOF
+
+  if ! diff $temp_file $enabled_file > /dev/null 2>&1
+  then
+    # Update the site file
+    mv $temp_file $site_file
+
+    # Enable the homestead nginx site
+    if ( nginx_ensite homestead > /dev/null )
+    then
+      service nginx stop
+    fi
+  else
+    rm $temp_file
+  fi
+fi
--- a/homestead.root/usr/share/clearwater/infrastructure/scripts/create-homestead-nginx-config
+++ b/homestead.root/usr/share/clearwater/infrastructure/scripts/create-homestead-nginx-config
@@ -38,11 +38,17 @@
 
 . /etc/clearwater/config
 
-if [ -n "$hs_hostname" ]
+split_domain_and_port()
+{
+  echo $1 | perl -ne '$_ =~ /(.+):([0-9]+)$/ && print "$1 $2\n"'
+}
+
+if [ -n "$hs_mgmt_hostname" ]
 then
   site_file=/etc/nginx/sites-available/homestead
   enabled_file=/etc/nginx/sites-enabled/homestead
   temp_file=$(mktemp homestead.nginx.XXXXXXXX)
+  server_name_and_port=( $( split_domain_and_port $hs_mgmt_hostname ) )
 
   cat > $temp_file << EOF
 upstream http_homestead {
@@ -54,7 +60,7 @@ upstream http_homestead {
 
 server {
         listen       [::]:8886 ipv6only=off;
-        server_name  $hs_hostname;
+        server_name  ${server_name_and_port[0]};
 
         location / {
                 proxy_pass http://http_homestead;
--- a/include/cache.h
+++ b/include/cache.h
@@ -597,7 +597,7 @@ public:
     ListImpus() {}
     virtual ~ListImpus() {}
 
-    std::vector<std::string>& get_impus_ref() { return _impus; }
+    virtual std::vector<std::string>& get_impus_ref() { return _impus; }
 
   protected:
     bool perform(CassandraStore::Client* client, SAS::TrailId trail);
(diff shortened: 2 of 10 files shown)


homestead.root/usr/share/clearwater/infrastructure/scripts/create-homestead-nginx-config
```diff
@@ -0,0 +1,89 @@
...
+  cat > $temp_file << EOF
+upstream http_homestead {
+        server unix:/tmp/homestead-http-mgmt-socket;
+
+        # The minimum number of idle connections to keep alive to the upstream.
+        keepalive 16;
+}
+
+server {
+        listen       [::]:8886 ipv6only=off;
```
graemerobertson: Shouldn't we listen on `server_name_and_port[1]`?

src/cache.cpp
```diff
@@ -995,3 +998,51 @@ bool Cache::DissociateImplicitRegistrationSetFromImpi::perform(CassandraStore::C
...
+  // In Cassandra, deleted rows appear like rows with no columns. So we need to
+  // ask for a column we know should exist (to determine if the row exists or
+  // not). The only column that must exists is the subscription XML, but this is
+  // large so we don't want to retrieve it for every subscriber. However, each
+  // row will have either been provisioned (meaning `_exists` will be present)
+  // or will have been created by registration (meaning `is_registered` will be
+  // present). So if we ask for both and get either back, then the row does
+  // really exists (and is not an artifact of a deleted row).
+  SlicePredicate sp;
+  sp.__set_column_names({IMS_SUB_XML_COLUMN_NAME});
```
graemerobertson: This comment needs updating :smile:

src/handlers.cpp
```diff
@@ -1325,7 +1325,7 @@ void ImpuRegDataTask::on_get_reg_data_success(CassandraStore::Operation* op)
       // Sprout wants to deregister this subscriber (because of a
       // REGISTER with Expires: 0, a timeout of all bindings, a failed
       // app server, etc.).
-      if (_original_state == RegistrationState::REGISTERED)
+      if (_original_state != RegistrationState::NOT_REGISTERED)
```
graemerobertson: I think this fixes an existing bug - good catch.

src/cache.cpp
```diff
@@ -995,3 +998,51 @@ bool Cache::DissociateImplicitRegistrationSetFromImpi::perform(CassandraStore::C
...
+  KeyRange kr;
+  kr.__set_start_key(""); // Start from the beginning.
+  kr.__set_end_key("");   // No end.
+  kr.__set_count(MAX_IMPUS_TO_RETURN);
+
+  client->get_range_slices(key_slices, cparent, sp, kr, ConsistencyLevel::ONE);
+
+  for (const KeySlice& key_slice: key_slices)
+  {
+    // Only report an IMPU if some of its columns exist (see comment above).
```
graemerobertson: This comment sort of needs updating too.

include/cache.h
```diff
@@ -587,6 +587,31 @@ class Cache : public CassandraStore::Store
...
+  /// that are assigned to this S-CSCF in the HSS).
+  /// - When subscribers are locally provisioned this returns all provisioned
+  /// subscribers.
+  class ListImpus : public CassandraStore::Operation
+  {
+  public:
+    ListImpus() {}
+    virtual ~ListImpus() {}
+
+    virtual std::vector<std::string>& get_impus_ref() { return _impus; }
```
graemerobertson: I'm sure this is a stupid question, but what is `ref`?

include/cache.h
```diff
@@ -587,6 +587,31 @@ class Cache : public CassandraStore::Store
...
     return new DissociateImplicitRegistrationSetFromImpi(impus, impis, timestamp);
   }
+
+  /// Get all the IMPUs for which Homestead has data in its cache.
+  ///
+  /// - When an HSS is in use, this lists all subscribers for which homestead is
+  /// storing data learned from the HSS (which in practise is all subscribers
+  /// that are assigned to this S-CSCF in the HSS).
+  /// - When subscribers are locally provisioned this returns all provisioned
+  /// subscribers.
```
graemerobertson: Worth mentioning that it only returns `MAX_IMPUS_TO_RETURN` or whatever?

src/handlers.cpp
```diff
@@ -2472,6 +2489,63 @@ void PushProfileTask::send_ppa(const std::string result_code)
...
+
+  rapidjson::StringBuffer sb;
+  rapidjson::Writer<rapidjson::StringBuffer> writer(sb);
+
+  writer.StartObject();
+  {
+    writer.String(JSON_IMPUS.c_str());
+    writer.StartArray();
+    {
+      for (const std::string& impu: impus)
```
graemerobertson: Hello 2016

src/main.cpp
```diff
@@ -645,15 +648,15 @@ int main(int argc, char**argv)
             options.sas_server,
             sas_write,
             options.sas_signaling_if ? create_connection_in_signaling_namespace
-                                     : create_connection_in_management_namespace);
+            : create_connection_in_management_namespace);
```
graemerobertson: Not sure I would have made this change!

src/ut/cache_test.cpp
```diff
@@ -1897,4 +1897,159 @@ TEST_F(CacheRequestTest, DissociateImplicitRegistrationSetFromWrongImpi)
...
+class ListImpusRecorder : public ResultRecorderInterface
+{
+public:
+  void save(CassandraStore::Operation* op)
+  {
+    impus = dynamic_cast<Cache::ListImpus*>(op)->get_impus_ref();
+  }
+  std::vector<std::string> impus;
+};
+
```
graemerobertson: Double blank line !!!! I'd move one of these blank lines up above `std::vector<std::string> impus;` FWIW.

src/ut/cache_test.cpp
```diff
@@ -1897,4 +1897,159 @@ TEST_F(CacheRequestTest, DissociateImplicitRegistrationSetFromWrongImpi)
...
+// Tests listing IMPUs when there aren't any in the database.
+TEST_F(CacheRequestTest, ListImpusNoResults)
+{
+  ListImpusRecorder rec;
+  RecordingTransaction* trx = make_rec_trx(&rec);
+  CassandraStore::Operation* op = _cache.create_ListImpus();
+
+  std::vector<cass::KeySlice> result;
+  std::vector<std::string> requested_columns = {"_exists", "is_registered"};
```
graemerobertson: As you probably already know, this needs changing.

src/ut/cache_test.cpp
```diff
@@ -1897,4 +1897,159 @@ TEST_F(CacheRequestTest, DissociateImplicitRegistrationSetFromWrongImpi)
...
+    .WillOnce(SetArgReferee<0>(result));
+  EXPECT_CALL(*trx, on_success(_))
+    .WillOnce(Invoke(trx, &RecordingTransaction::record_result));
+
+  execute_trx(op, trx);
+
+  EXPECT_TRUE(rec.impus.empty());
+}
+
+// Tests listing IMPUs when there there is on IMPU, that has been dynamically
```
graemerobertson: *one

src/ut/cache_test.cpp
```diff
@@ -1897,4 +1897,159 @@ TEST_F(CacheRequestTest, DissociateImplicitRegistrationSetFromWrongImpi)
...
+TEST_F(CacheRequestTest, ListImpusOneResultRegistered)
+{
+  ListImpusRecorder rec;
+  RecordingTransaction* trx = make_rec_trx(&rec);
+  CassandraStore::Operation* op = _cache.create_ListImpus();
+
+  std::vector<cass::KeySlice> result(1);
+  result[0].key = "Kermit";
+  // Set the "is_registered" column to make the subscriber appear registered.
+  make_slice(result[0].columns, {{"is_registered", "\x01"}});
```
graemerobertson: Again, this is no longer a thing.

src/ut/cache_test.cpp
```diff
@@ -1897,4 +1897,159 @@ TEST_F(CacheRequestTest, DissociateImplicitRegistrationSetFromWrongImpi)
...
+  EXPECT_CALL(*trx, on_success(_))
+    .WillOnce(Invoke(trx, &RecordingTransaction::record_result));
+
+  execute_trx(op, trx);
+
+  EXPECT_EQ(1, rec.impus.size());
+  EXPECT_EQ("Kermit", rec.impus[0]);
+}
+
+// Tests listing IMPUs when there there is on IMPU, that has been configured but
```
graemerobertson: *one

src/ut/cache_test.cpp
```diff
@@ -1897,4 +1897,159 @@ TEST_F(CacheRequestTest, DissociateImplicitRegistrationSetFromWrongImpi)
...
+TEST_F(CacheRequestTest, ListImpusOneResultConfigured)
+{
+  ListImpusRecorder rec;
+  RecordingTransaction* trx = make_rec_trx(&rec);
+  CassandraStore::Operation* op = _cache.create_ListImpus();
+
+  std::vector<cass::KeySlice> result(1);
+  result[0].key = "Kermit";
+  // Set the _exists column to show the subscriber is configured.
+  make_slice(result[0].columns, {{"_exists", ""}});
```
graemerobertson: The distinction between this test and the previous one is probably not relevant any more. Anyway, I'm going to stop commenting on this and let you sort it out.

src/main.cpp
AlexHockey: That was how Vim decided to reformat things. I'll put it back.

include/cache.h
AlexHockey: It's to make it clear that this function returns a reference to the impus vector (rather than returning it by value and doing a lot of unnecessary copying). I'll spell it out.

homestead.root/usr/share/clearwater/infrastructure/scripts/create-homestead-nginx-config
AlexHockey: I wondered about that, but we don't do it for any other interfaces (e.g. homestead does not listen on the port specified in `hs_hostname`).

graemerobertson: Isn't that just a bug though?

It feels like currently `hs_mgmt_hostname` needs to be set to `<host>:8886`, right? That's rather odd...

include/cache.h
graemerobertson: Ah ok. Should the argument by `const` then? I'd call the function `get_impus_by_ref` but I see you've already changed it to `get_impus_reference`. Anyway, that's fine.

AlexHockey:
@rkd-msw PR for cached data API

homestead.root/usr/share/clearwater/infrastructure/scripts/create-homestead-nginx-config
AlexHockey: It's consistent though. If we want to change our products to respect the port specified in /etc/clearwater/shared_config I think we should do them all at once, otherwise it's confusing for us and for support.

PR #377, opened 2016-10-07
[Reviewer: Graeme] Allow removal of user assigned but not registered

rkd-msw:


graemerobertson:
As discussed, I'm going to make this change on my branch because it'll make code management easier

PR #378, opened 2016-10-11
[Reviewer: Graeme] Follow-ons from #375

AlexHockey:
Adds some non-essential code as a follow-on from #375:

* Adds a unit test for one of the bugs found during live test. 
* Changes a function to return a const reference for safety.


graemerobertson:
Love it!

The infratstructure in `handlers_test.cpp` is paying for itself!

PR #381, opened 2016-10-24
[Reviewer: Sathiyan] Add homestead-cx package

eleanor-merry:


sathiyan-sivathas:
Do you need something in the Makefile? Like this:
https://github.com/Metaswitch/homestead/blob/dev/Makefile#L16

I don't know if it's strictly necessary though.

PR #383, opened 2016-11-03
[Reviewer AJH] Implement cx poll script

steelyvoid:
This PR adds the facility to have monit check the status of the Cx interface as seen by homestead, and to restart homestead if all UAAs, MARs, SARs and LIRs result in more than a configured threshold (default 20 per 5 minute interval) of server errors (i.e. errors leading to 5XX errors at Homestead's HTTP interface) with no successes being seen over the same period.

The motivation for this enhancement is to recover Homestead in scenarios where unexpected errors in VM networking result in a hung Diameter stack which persistently failed all Diameter requests even when the underlying VM networking problems were resolved.  In the case seen, the Diameter stack had apparently entered a protocol level deadlock with the HSS - restarting homestead fixed the problem.

This has been tested using iptables rules to artificially break the connectivity to the HSS (such that all Diameter requests failed) and then registering to force suitable 5XX-generating Diameter response codes to be generated.  Numbers of errors both above and below the threshold were tested and the restart behaviour (restart if above, otherwise don't restart) was confirmed.

The need for the "...for 2 cycles" clause in homestead.monit comes from the fact that monit effectively reads poll script returncodes one cycle late (so if we restarted as soon as we saw a poll_cx failure, we would effectively restart homestead twice).  I think this is a known limitation of monit since all our existing poll scripts also use 2 cycles.

One point for discussion:  the poll_cx.py and handlers.cpp files are implicitly linked, in that, should the mapping of Diameter response codes to HTTP codes in the latter be changed, it will also need to be changed in the former.  Any ideas for making this clear to future developers?  On the face of it, this could be a maintainability headache.

Uses the stop_or_abort script introduced by https://github.com/Metaswitch/clearwater-infrastructure/pull/392


rkd-msw:
Two thoughts from me:

- why is this in a Monit script and not in homestead itself? There's precedent for detecting bad state in product code and aborting (https://github.com/Metaswitch/sprout/blob/287d08db274b294a9f9f9559e6124c8975db64b7/src/thread_dispatcher.cpp#L228) and I think doing so would remove the Diameter-to-HTTP mapping issue.
- should we be using the stop_or_abort script everywhere (or making an equivalent Monit enhancement)?

steelyvoid:
My original plan here (discussed with Alex) was to have a generic mechanism for spotting when Homestead was, or might be, in some "bad state", and restart it in an attempt to clear the problem.  An internal mechanism would have had to be pretty smart to make that determination, but something that looked at stats and could make the determination externally could be relatively simple in its algorithms (if we hit "server" errors and no successes over some time period, we might be borked)

That then became a plan to restart based on Cx errors only when I realised that you can't look at Homestead's own response codes to determine whether, say, the freeDiameter stack was broken, because such a homestead could still return success on any request that hit Cassandra only.

My gut feel is that its still better to have external mechanisms for liveness (and monit is the "resilience compoent", so is the logical place for such mechanisms), but it does mean that we have the issue of needing knowledge about what are Cx "Server error/client errors/successes" in two places.

Alex, what do you think?  One way to resolve the dilemma is to go with your original suggestion that we blacklist particular Cx errors and say "If we see non-zero numbers of these, and zero numbers of anything else, we are potentially borked".  In my testing, where I explicitly broke the connectivity to the HSS, I was getting 3002 errors (DIAMETER_UNABLE_TO_DELIVER).  Perhaps we should just change the logic to say "If I see non-zero numbers of 3002s (across all tables) plus timeouts, and zero instances of any other return codes, we are borked", and then if L3 see any instances in the field that aren't covered by this, we have the option of tweaking that logic.

Rob, on the stop_or_abort script, Alex and I felt we should stick with using this just in this instance for now and potentially expand its use in future if it turns out to be a useful idea.  I guess one option would be to allow it to take a "reason string" (set to "poll_cx" in this case) which could then be tacked onto the end of the marker file, and then we could change all existing monit poll scripts to use it (so all poll failures can general core files independently of one another, but only 1 per day for any given poll script.  Thoughts, Alex?

rkd-msw:
I suspect that actually, external means of detecting this would be more complicated than internal ones - the information available to an external check is generally going to be a subset of what's available to an internal check (because all the externally-visible information, like stats, is generated internally). The point you make about not knowing externally whether Homestead's HTTP successes reflect a Diameter success or just successful retrieval of cached data is an example of this, I think - but I don't feel too strongly so I'm happy to drop this line of argument if you disagree.

Fine on `stop_or_abort`, but I think we should be tracking somewhere that this is a work in progress (I'm worried that otherwise we'll forget it, and we'll hit unusual issues in Homestead because it's using a different script from everything else).

homestead.root/usr/share/clearwater/bin/poll_cx.py
```diff
@@ -0,0 +1,287 @@
...
+
+# Get tolerance to server errors. We will only treat the number of server
+# errors at the Diameter interface as serious enough to warrant a restart
+# (provided there have been no successes over the same period) if the total
+# number of errors exceeds this tolerance.
+server_error_tolerance = int(sys.argv[1])
+_log.info("Server error tolerance = {}".format(server_error_tolerance))
+
+# Get a connection to the local SNMP agent
+ss = netsnmp.Session(DestHost='localhost', Version=2, Community='clearwater')
```
rkd-msw: This should be `clearwater-internal` (see [here](https://github.com/Metaswitch/clearwater-infrastructure/blob/6ffb0f6a021b78845583ce2d1e6b882becdcec6d/clearwater-snmpd/etc/snmp/snmpd.conf.clearwater-snmpd#L247)) - for security, users can change the community string that's accessible from other boxes from `clearwater` to something else.

AlexHockey:
I think the approach of using stats (and other general indicators) to monitor the health of a process is a good one. We have one microservice that provides the business logic of homestead, and one microservice that ensures the first is running properly, with a well-defined API between them. Each microservice focuses on one set of things, and is coded appropriately - the business logic is in C++ for speed, and the monitoring is in bash/python for flexibility. This feels like a clean, composable and extensible architecture to me!

Looking in this PR specifically, suppose there are some bugs and homestead gets restarted when it shouldn't, or doesn't when it should. In this case someone could take corrective action like disabling the poll_cx job in monit, or adding extra diameter codes to treat as errors to the python script. This is much easier than recompiling and patching homestead. 

On the topic of coupling poll_cx and handlers.cpp together, I am less concerned about this:

1. I view the poll_cx as simply knowing what diameter return codes point to a broken diameter connection, and that we have determined this by looking at what handlers.cpp does (as a shortcut to reading and understanding all the return codes, and the code paths through freeDiameter, and compiling a list that way). I could imagine a future where we find error codes that we want to exclude from the check that nevertheless result in a 5xx HTTP response, so I don't think the handler.cpp -> poll_cx mapping should be rigid.

2. handler.cpp is massive. We should be breaking it up to have one file per handler. If we did that - where would the logic that determines whether to restart homestead live? It would probably live in a separate file, away from the handlers themselves. So we would have the same maintainability problem.

On `stop_or_abort` I think it would be good to make this generic, but not to change any existing monit jobs right now. @steelyvoid can you make that change and, when the PR is merged send round a PSA to the PC team? I will add updating other monit jobs to our tech debt list.

AlexHockey:
One more thing: I'm planning to review this tomorrow afternoon - @steelyvoid hopefully you don't need it turned around before then?

steelyvoid:
Thanks Alex.  No great rush

src/handlers.cpp
```diff
@@ -343,14 +343,6 @@ void ImpiTask::on_mar_response(Diameter::Message& rsp)
...
   _maa->experimental_result(experimental_result_code, vendor_id);
 
-  if (result_code != 0)
-  {
-    mar_results_tbl->increment(SNMP::DiameterAppId::BASE, result_code);
-  }
-  else if (experimental_result_code != 0 && vendor_id == VENDOR_ID_3GPP)
-  {
-    mar_results_tbl->increment(SNMP::DiameterAppId::_3GPP, experimental_result_code);
-  }
```
AlexHockey: I understand from @bossmc that IMS mandates that responses have exactly one of these, so the code here is actually fine. It's probably worth re-instating it given it's a fair bit simpler than the other option.

steelyvoid:
Thanks Alex

In your general comments above, you suggest categorising responses as either

- "failures" (non-2001 result codes)
- "successes" (2001 result codes or any experimental code)

My original code made a distinction between client errors and server errors, given that, if the only errors you're getting are "authorization" errors (for instance), you don't have any evidence that the server connection itself is broken. Making the above change would mean that if, say, we had a stream of "DIAMETER_AUTHORIZATION_REJECTED" result codes on UARs (and nothing else), we'll do a "potentially broken" homestead restart, even thought homestead itself regards this as a client error (it returns 403 Forbidden).

Is the answer to treat certain other non-2001 result codes as "successes" for the purposes of this check? Perhaps  DIAMETER_AUTHORIZATION_REJECTED (4001) and DIAMETER_LIMITED_SUCCESS (2002) are the only returncodes that we can treat as successes - the only other 2XXX/4XXX returncode is DIAMETER_OUT_OF_SPACE (4003), which sounds a bit "servery" as opposed to "clienty" (but the explanation for it is "A Diameter node received the accounting request but was unable to commit it to stable storage due to a temporary lack of space" which suggests that you can't get it unless the server is contactable).

What if we were to say that any 2XXX or 4XXX resultcode (or any experimental code) is a "success" and anything else is a "failure"?

AlexHockey:
I was trying to match your existing implemention of poll_cx, but I see that sometimes you do treat 2002 as a success. I like your suggestion of treating 2XXX and 4XXX as success and other errors as "failures". Just to be clear I think the end result will be:

* Response with *experimental* result code => success
* Response with 2XXX or 4XXX result code => success
* Other response => failure.

steelyvoid:
Agreed - thanks

homestead.root/usr/share/clearwater/bin/poll_cx.py
```diff
@@ -0,0 +1,287 @@
...
+                          count_responses(mar_segment_head + ".1", [], [5001]))
+
+    # Add the count of timeouts to the server error count
+    total_server_errors += count_timeouts(mar_segment_head)
+
+    _log.debug("Total MAR successes = {}".format(total_successes))
+    _log.debug("Total MAR client errors = {}".format(total_client_errors))
+    _log.debug("Total MAR server errors = {}".format(total_server_errors))
+
+    return (total_successes, total_client_errors, total_server_errors)
```
steelyvoid: Will do.  Code is much simpler now

homestead.root/usr/share/clearwater/bin/poll_cx.py
```diff
@@ -0,0 +1,287 @@
...
+#
+# Note that this is called twice to ensure that we do not make a false
+# positive determination due to "tick rollover"
+def are_all_requests_failing():
+
+    # Collate all "Previous 5 minute" statistics for the UAR, MAR, SAR and LIR
+    # statistic tables
+    (total_successes, total_client_errors, total_server_errors) = \
+              map(operator.add, map(operator.add, MAR_responses(), SAR_responses()), \
+                                map(operator.add, UAR_responses(), LIR_responses()))
```
steelyvoid: Actually, now that I've got rid of the command specific collators, map(operator.add makes more sense (because I'm just calling the same routine repeatedly maintaining a running count)

src/ut/handlers_test.cpp
```diff
@@ -2916,56 +2977,9 @@ TEST_F(HandlersTest, IMSSubscriptionUserUnknownDeregCacheFail)
...
-
-  EXPECT_CALL(*_httpstack, send_reply(_, 500, _));
-  _caught_diam_tsx->on_response(saa);
-
-  delete _caught_diam_tsx; _caught_diam_tsx = NULL;
-  _caught_fd_msg = NULL;
+  sa_error_template(0, 0, 500);
+  sa_error_template(DIAMETER_AUTHORIZATION_REJECTED, 0, 500);
+  sa_error_template(0, DIAMETER_ERROR_ROAMING_NOT_ALLOWED, 500);
 }
```
steelyvoid: Thanks.  Actually, now that I've reverted changes to handlers.cpp (apart from comments underlining the fact that there will only be at most one of result code or experimental code at any given time) I don't need these test changes, so I'll back these out as well

src/handlers.cpp
steelyvoid: Yep - will do.  Thanks

homestead.root/usr/share/clearwater/bin/poll_cx.py
```diff
@@ -0,0 +1,287 @@
...
+# INPUT
+#   head_oid         - Pre-indexed OID of the segment of the table we are
+#                      searching. This will be the OID of the Cx response
+#                      table concatenated with:
+#                      -  ".1.4" (".1" says "entry" and ".4" says "count element")
+#                      -  ".3" ("Previous 5 minute values")
+#                      -  AppID.  This will be ".0" for "Base" and ".1" for _3GPP
+#   successes        - The response codes in this segment that correspond to
+#                      2XX "success" HTTP codes from a Homestead user's
+#                      perspective
```
steelyvoid: Understood.  I've changed as per our agreed strategy below (all experimental and 2XXX and 4XXX result codes mean success, anything else means potential connection failure)

homestead.root/usr/share/clearwater/bin/poll_cx
```diff
@@ -0,0 +1,45 @@
...
+. /etc/clearwater/config
+
+output=$(/usr/share/clearwater/bin/poll_cx.py "${hss_server_error_tolerance:-20}")
+
+if [ "$output" = "All requests failing? True" ]
+then
+  exit 3
+else
+  exit 0
+fi
```
steelyvoid: Will do

steelyvoid:
Hi Alex.

I've made and retested those simplifications/changes.  Can you cast another eye over (and look at the linked PR in clearwater-infrastructure)?

Thanks!

AlexHockey:
Sweet.

Issue #389, opened 2016-11-23
Homestead returns inconsistent HTTP errors (500 and 503) for the same underlying error (THRIFT_EAGAIN)


rkd-msw:
#### Symptoms

This is the same issue described in #387 (REGISTERs failing at load due to a Cassandra database error). This issue covers the fact that it sometimes causes a 500 HTTP error and sometimes a 503.

Specifically:

* if Homestead hits this error when reading from Cassandra it sends a 503
* if Homestead hits this error when writing to Cassandra it sends a 500

#### Impact

Sending a 500 instead of a 503 has an impact on call service - specifically, a 500 error doesn't get retried to a different Homestead, so REGISTERs will fail when they might have succeeded on a retry.

There's also a diagnostic impact - having different error codes means that it's harder to spot that these are the same underlying issue and means we're more likely to spend time investigating duplicate bugs.

#### Release and environment 

CC-homestead-10.0.0-20161116-~255.00.0.ova

#### Steps to reproduce

View the two traces at \\ad\csd\bigfiles\clearwater\issue-attachments\homestead\389, which I got by reproducing #387 (i.e. by running moderate levels of REGISTER stress).

eleanor-merry:
Passed back for retest

rkd-msw:
Can't easily retest because #387 is fixed - I'm going to close it on the grounds that your fix is UTed and that's good enough.

PR #391, opened 2016-11-29
[Reviewer: Seb] Make error codes consistent

eleanor-merry:
Consistently return a 503 when we hit an error on a request that can be retried.

Tested in UT

Fixes issue #389


sebrexmetaswitch:
LGTM

PR #393, opened 2016-12-01
[Reviewer: Ellie] Use loopback IP rather than localhost for default cassandra hostname

### LINKED ISSUE #386: Homestead requires a DNS entry for localhost
#### Symptoms
I spun up a split-network Homestead node. The Homestead process was unable to resolve localhost and so couldn't connect to Cassandra.

This works in a single network system because dnsmasq is installed in the management namespace and dnsmasq uses /etc/hosts/ which contains an entry for localhost. Similarly, this would work in a split network system where the signaling DNS server had an entry for localhost. Another work around is to set cassandra_hostname to be 127.0.0.1. However, these workarounds are not documented anywhere as far as I can tell, and it’s not at all clear to me that they’re the right thing to be doing.

#### Impact
Nothing worked.

#### Release and environment 
CC-homestead-10.0.0-20161116-~255.00.0.ova

#### Steps to reproduce
Spin up node

sebrexmetaswitch:
Fixes #386 

Testing done:
- live tested (that not specifying a cassandra hostname uses 127.0.0.1)
- UTs and live tests pass

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -192,7 +192,7 @@ void usage(void)
        " -H, --http <address>       Set HTTP bind address (default: 0.0.0.0)\n"
        " -t, --http-threads N       Number of HTTP threads (default: 1)\n"
        " -u, --cache-threads N      Number of cache threads (default: 10)\n"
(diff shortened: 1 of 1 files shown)


sebrexmetaswitch:
(Ellie gave this a quick look)

PR #394, opened 2016-12-07
[Reviewer: Ellie] Suppress output of poll_cassandra when adding schemas

### LINKED ISSUE #333: Errors reported when upgrading the Clearwater packages
#### Symptoms

When upgrading the dogfood system I logged in to Homestead and ran sudo clearwater-upgrade. This was successful however lots of connection errors were reported like:

Connection error: Could not connect to localhost:9160
#### Impact

It made me slightly nervous. 
#### Release and environment

Homestead node on the VMware deployment.

sebrexmetaswitch:
We already print an error and exit if we can't poll cassandra within 2 minutes,
so this removes a benign but scary sounding log when you upgrade

Fixes #333 

**Testing done**
 - tested that we no longer see the benign logs when upgrading

### DIFF
--- a/homestead-cassandra.root/usr/share/clearwater/cassandra-schemas/homestead_cache.sh
+++ b/homestead-cassandra.root/usr/share/clearwater/cassandra-schemas/homestead_cache.sh
@@ -8,6 +8,8 @@ cassandra_hostname="127.0.0.1"
 
 quit_if_no_cassandra
 
+echo "Adding Cassandra schemas..."
+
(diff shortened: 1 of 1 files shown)


sebrexmetaswitch:
I also had a look, and I think we're upgrading the clearwater-cassandra, homestead-cassandra and homestead packages in the correct order, so I think the solution here is just to suppress the output of the poll_cassandra.sh script while we're trying to add the schemas

PR #398, opened 2016-12-07
[Reviewer: Ellie] Add SAS log for unsupported auth scheme from HSS

### LINKED ISSUE #308: Homestead provides no useful indication why it fails for a MAA with an unsupported authentication scheme
Homestead does not provide a useful log (at level 5) indicating why it returns a 404 when receiving an unsupported scheme in a SIP-Authentication-Scheme AVP of an MAA. The logfile contains the following:

```
04-03-2016 16:30:53.905 UTC Debug handlers.cpp:336: Received Multimedia-Auth answer with result code 2001
04-03-2016 16:30:53.905 UTC Debug cx.cpp:459: Got SIP-Auth-Scheme Digest
04-03-2016 16:30:53.905 UTC Verbose httpstack.cpp:69: Sending response 404 to request for URL /impi/3196910004%40telcocloud.com/av, args impu=sip%3A3196910004%40telcocloud.com
```

It could be inferred that the 404 is probably due to getting a Digest scheme, but this is not clear (the cx.cpp trace is informational, on_mar_response() in heandlers.cpp has no trace for when a scheme does not match SIP Digest or AKA). More importantly, there is no SAS log generated indicating why the MAR/MAA exchange fails.

sebrexmetaswitch:
Fixes #308 

**Testing done**
 - tested that an unsupported scheme (in this case `Early-IMS-Security`) triggered this debug log and SAS log

### DIFF
--- a/include/homesteadsasevent.h
+++ b/include/homesteadsasevent.h
@@ -45,6 +45,7 @@ namespace SASEvent
   // Homestead events
   //----------------------------------------------------------------------------
   const int INVALID_SCHEME = HOMESTEAD_BASE + 0x0000;
+  const int UNSUPPORTED_SCHEME = HOMESTEAD_BASE + 0x0001;
   const int NO_IMPU_AKA = HOMESTEAD_BASE + 0x0010;
   const int NO_AV_CACHE = HOMESTEAD_BASE + 0x0020;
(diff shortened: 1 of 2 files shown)


eleanor-merry:
LGTM

PR #417, opened 2017-01-11
[Reviewer: Seb] Don't create schemas on infrastructure restart

eleanor-merry:
I'll also remove all the other schema creation files from this location (in the other repos).


sebrexmetaswitch:
Looks good.

Presumably we also need to update readthedocs to explain that when not using etcd you need to manually run the /usr/share/clearwater/cassandra-schemas/homestead_cache.sh (and memento, homer etc.) scripts, since these won't now be run automatically.

eleanor-merry:
I'll update the docs

PR #422, opened 2017-01-24
[Reviewer: Matt] Send HTTP 503 rather than 504 on diameter transaction timeout

sebrexmetaswitch:
This means we treat a timeout as the same as having no diameter connections, and
allows the HTTP client to retry the request to another node if appropriate.

This is intended as a fix for https://github.com/Metaswitch/sprout/issues/1658

Testing done:
 - manually verified that blocking HSS connections from one homestead now results in a 503 which is retried by the S-CSCF to another homestead node, resulting in the registration succeeding
 - fixed UTs and verified they all pass
 - the live tests pass


mirw:
Looks good.

PR #435, opened 2017-03-08
[Reviewer: Krista] Wildcard support

eleanor-merry:
This adds support for Homestead to pass wildcard information to/from Sprout.

Tested in UTs and live.


docs/homestead_api.md
```diff
@@ -82,6 +82,8 @@ The valid values of reqtype are:
 
 The body also has an optional `server_name` field, specifying the S-CSCF URI to use in the Server-Name on the Server-Assignment-Request. If this is missing then the Server-Name is set to the value of the configured S-CSCF URI (in `/etc/clearwater/shared_config`).
 
+The body also has another optional `wildcard-identity` field, specifying what wildcarded public user identity is associated with the public ID. This is only used by Homestead if the `reqtype` is either `call` or `dereg-user`, and it is used in the Wildcarded-Public-Identity AVP in the Server-Assignment-Request. If this field is missing then the Wildcarded-Public-Identity AVP is not set.
```
KristaHyer: Isn't it used in more that just 'call' and 'dereg-user' cases?

docs/homestead_api.md
```diff
@@ -82,6 +82,8 @@ The valid values of reqtype are:
...
+
```
KristaHyer: isn't it used in more than just the 'call' and 'dereg-user' cases?

src/ut/handlers_test.cpp
```diff
@@ -95,6 +95,7 @@ class HandlersTest : public testing::Test
   static const std::string DEST_HOST;
   static const std::string DEFAULT_SERVER_NAME;
   static const std::string SERVER_NAME;
+  static const std::string WILDCARD;
```
KristaHyer: Change from "wildcard" to "wildcarded_public_identity" - same change needed in lots of places.

src/ut/handlers_test.cpp
```diff
@@ -3497,7 +3520,58 @@ TEST_F(HandlersTest, LocationInfo)
...
+  // Turn the caught Diameter msg structure into a LIR and check its contents.
+  Diameter::Message msg(_cx_dict, _caught_fd_msg, _mock_stack);
+  Cx::LocationInfoRequest lir(msg);
+  EXPECT_TRUE(lir.get_str_from_avp(_cx_dict->DESTINATION_REALM, test_str));
+  EXPECT_EQ(DEST_REALM, test_str);
+  EXPECT_TRUE(lir.get_str_from_avp(_cx_dict->DESTINATION_HOST, test_str));
+  EXPECT_EQ(DEST_HOST, test_str);
+  EXPECT_EQ(IMPU, lir.impu());
+  EXPECT_FALSE(lir.originating(test_i32));
+  EXPECT_FALSE(lir.auth_type(test_i32));
```
graemerobertson: I know this is exactly the same as all the other tests in this file (most of which I wrote...) but one of the main problems with these UTs is that they all test everything for no reason - none of these checks are useful; they're all tested by other tests. This test could/should be a lot more focused on what it's actually testing.

docs/homestead_api.md
eleanor-merry: I don't think so - these two cases get translated into the six SAR types.

src/handlers.cpp
```diff
@@ -1120,6 +1138,24 @@ std::string ImpuRegDataTask::server_name_from_body(std::string body)
   }
 }
 
+std::string ImpuRegDataTask::wildcard_from_body(std::string body)
+{
+  rapidjson::Document document;
+  document.Parse<0>(body.c_str());
+
+  if (!document.IsObject() ||
+      !document.HasMember("wildcard-identity") ||
```
eleanor-merry: I've made the -/_ consistent within a request (but they are different for LIRs/SARs)

docs/homestead_api.md
KristaHyer: ah ok, this makes sense then

PR #437, opened 2017-03-09
Support wildcarded public identities on Homestead/HSS interface

KristaHyer:
Adding capabilities to:
 Put wildcard on SAR
 Search cache with wildcard
 Read wildcard off SAA and search cache with that

Some testing has been done - to be completed by EM.


src/handlers.cpp
```diff
@@ -1893,6 +1933,25 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
   return;
 }
 
+std::string ImpuRegDataTask::search_term()
```
eleanor-merry: I think this should be two functions.
The first one (wildcard_id()?) covers "_hss_wildcard.empty() ? _sprout_wildcard : _hss_wildcard". 
The second (public_id()) then can use this - e.g. std::string wildcard_id = wildcard_id(); return wildcard_id.empty() ? _impu : wildcard_id;"

src/handlers.cpp
```diff
@@ -1818,6 +1823,41 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
     SAS::report_event(event);
     _http_rc = HTTP_NOT_FOUND;
   }
+  // An SAA received with this experimental result code may be updating the
+  // wildcarded public identity. If it is, then return to searching the cache
+  // with the new wildcarded public identity and proceed from there. But if it
+  // isn't, treat it as an error.
+  else if (experimental_result_code == DIAMETER_ERROR_IN_ASSIGNMENT_TYPE)
+  {
+    const std::string current_wildcard = (_hss_wildcard.empty() ? _sprout_wildcard : _hss_wildcard);
```
eleanor-merry: Line length?

include/handlers.h
```diff
@@ -522,7 +526,8 @@ class ImpuRegDataTask : public HssCacheTask
   ChargingAddresses _charging_addrs;
   long _http_rc;
   std::string _provided_server_name;
-  std::string _wildcard_identity;
```
eleanor-merry: Can you comment what these variables are? (e.g. why is it important to have both?)

src/handlers.cpp
```diff
@@ -1818,6 +1823,41 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
...
+    saa.wildcarded_public_identity(_hss_wildcard);
+    if (current_wildcard == _hss_wildcard)
+    {
+      int type = 0;
+      saa.server_assignment_type(type);
+      TRC_INFO("Server-Assignment answer with result code %d and experimental "
```
eleanor-merry: Can the trace statement also include the current_wildcard? (We're not saying that this is definitely a wildcard issue here, but including it in the trace makes it quicker to debug).

src/handlers.cpp
```diff
@@ -1818,6 +1823,41 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
...
+      SAS::Event event(this->trail(), SASEvent::REG_DATA_HSS_FAIL_ASSIGNMENT_TYPE, 0);
+      event.add_static_param(result_code);
+      event.add_static_param(experimental_result_code);
+      event.add_static_param(type);
+      SAS::report_event(event);
+      _http_rc = HTTP_SERVER_ERROR;
+    }
+    else
+    {
+      // Return to search the cache with the new wildcarded public identity, and
```
eleanor-merry: The comment is a bit terse - possibly send another SAR, send a HTTP response,...

src/handlers.cpp
```diff
@@ -1818,6 +1823,41 @@ void ImpuRegDataTask::on_sar_response(Diameter::Message& rsp)
...
+    if (current_wildcard == _hss_wildcard)
+    {
+      int type = 0;
+      saa.server_assignment_type(type);
+      TRC_INFO("Server-Assignment answer with result code %d and experimental "
+               "result code DIAMETER_ERROR_IN_ASSIGNMENT_TYPE, with vendor id "
+               "%d and assignment type %d - reject",
+               result_code, vendor_id, type);
+      SAS::Event event(this->trail(), SASEvent::REG_DATA_HSS_FAIL_ASSIGNMENT_TYPE, 0);
+      event.add_static_param(result_code);
```
eleanor-merry: The result code isn't interesting here (as we've already decided that we're interested in the experimental result code).

src/handlers.cpp
eleanor-merry: This doesn't need to be const here.

PR #445, opened 2017-04-26
Support barred IMPUs for RTRs

KristaHyer:
We can no longer guarantee that the default public id is the first public id in a registration set.
Instead, it is the first UNBARRED public id in a registration set.

Extra logic to pull out this updated definition of the default id has been added to the code that supports RTRs.

Several tests have been added to double check this works.
More tests need to be added in the future, before this code is added to V11, to handle the full range of RTRs that can be recieved.

### DIFF
--- a/src/homestead_xml_utils.cpp
+++ b/src/homestead_xml_utils.cpp
@@ -222,8 +222,17 @@ std::vector<std::string> get_public_ids(const std::string& user_data)
   return get_public_and_default_ids(user_data, unused_default_id);
 }
 
+// Parses the given User-Data XML to retrieve the default public ID (the first
+// unbarred public ID).
+void get_default_id(const std::string& user_data,
+                    std::string &default_id)
+{
+  std::vector<std::string> unused_public_ids;
+  unused_public_ids = get_public_and_default_ids(user_data, default_id);
+}
+
 // Parses the given User-Data XML to retrieve a list of all the public IDs, and
-// the default id (the first unbarred public ID).
+// the default public ID (the first unbarred public ID).
 std::vector<std::string> get_public_and_default_ids(const std::string &user_data,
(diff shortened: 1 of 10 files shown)


src/ut/handlers_test.cpp
```diff
@@ -1623,10 +1706,13 @@ std::vector<std::string> HandlersTest::IMPU_IN_VECTOR = {IMPU};
 std::vector<std::string> HandlersTest::IMPI_IN_VECTOR = {IMPI};
 std::vector<std::string> HandlersTest::ASSOCIATED_IDENTITY1_IN_VECTOR = {ASSOCIATED_IDENTITY1};
 std::vector<std::string> HandlersTest::IMPU_REG_SET = {IMPU, IMPU4};
+std::vector<std::string> HandlersTest::IMPU_REG_SET2 = {IMPU, IMPU2};
```
sathiyan-sivathas: Actually these are both reg sets for IMPU 1 so this is the right naming?

src/homestead_xml_utils.cpp
```diff
@@ -280,6 +303,13 @@ std::vector<std::string> get_public_ids(const std::string& user_data)
   {
     TRC_ERROR("Failed to extract any ServiceProfile/PublicIdentity/Identity nodes from %s", user_data.c_str());
   }
+
+  // Set the default id - this is the first unbarred public identity.
+  if (unbarred_public_ids.size() != 0)
```
sathiyan-sivathas: I would have done `!unbarred_public_ids.empty()` but the existing code does the same thing for `public_ids` just above so I don't really care.

src/ut/handlers_test.cpp
KristaHyer: Yh that's what I figured too.

PR #446, opened 2017-05-04
[Reviewer: KH1] Add support for the UAR-Flags AVP

sathiyan-sivathas:
This PR takes the "sos=true" parameter on the UAR HTTP GET request from the I-CSCF and adds the UAR-Flags AVP on the Diameter UAR. Everything should be as in the design.

Sprout PR will follow shortly.

### DIFF
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -623,6 +623,8 @@ void ImpiRegistrationStatusTask::run()
       _visited_network = _dest_realm;
     }
     _authorization_type = _req.param("auth-type");
+    std::string sos = _req.param("sos");
+    _emergency = sos == "true" ? true : false;
     TRC_DEBUG("Parsed HTTP request: private ID %s, public ID %s, visited network %s, authorization type %s",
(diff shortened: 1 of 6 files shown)


src/handlers.cpp
```diff
@@ -623,6 +623,8 @@ void ImpiRegistrationStatusTask::run()
       _visited_network = _dest_realm;
     }
     _authorization_type = _req.param("auth-type");
+    std::string sos = _req.param("sos");
+    _emergency = sos == "true" ? true : false;
     TRC_DEBUG("Parsed HTTP request: private ID %s, public ID %s, visited network %s, authorization type %s",
```
KristaHyer: Should "sos" and "true" be in a header, as they are just strings?
Also, since these are passed from sprout, shouldn't they be in a shared header so that we don't change them in one and break everything?

Although, the "auth-type" just above them is not, so if strings are used in the rest of this file, I'd be happy for you to leave these new strings you've added.

sathiyan-sivathas: I'll leave this as is because the rest of the file deals with strings and not constants. I like your idea though.

PR #448, opened 2017-05-05
[Reviewer: Krista] Add config option to control shared IFC support

eleanor-merry:
Adds support for a new config option to control setting the supported-features AVP for shared IFC sets

### DIFF
--- a/debian/homestead.init.d
+++ b/debian/homestead.init.d
@@ -160,6 +160,7 @@ get_daemon_args()
 
 
         [ "$sas_use_signaling_interface" != "Y" ] || sas_signaling_if_arg="--sas-use-signaling-interface"
+        [ "$support_shared_ifc" == "N" ] || support_shared_ifcs_arg="--support-shared-ifcs"
 
         [ -z "$diameter_timeout_ms" ] || diameter_timeout_ms_arg="--diameter-timeout-ms=$diameter_timeout_ms"
         [ -z "$signaling_namespace" ] || namespace_prefix="ip netns exec $signaling_namespace"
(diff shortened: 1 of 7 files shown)


src/main.cpp
```diff
@@ -174,6 +176,7 @@ const static struct option long_opt[] =
   {"pidfile",                     required_argument, NULL, PIDFILE},
   {"daemon",                      no_argument,       NULL, DAEMON},
   {"sas-use-signaling-interface", no_argument,       NULL, SAS_USE_SIGNALING_IF},
+  {"support-shared-ifcs",         no_argument,       NULL, SUPPORT_SHARED_IFCS},
```
KristaHyer: I'm slightly confused,  this is the config option, right?
Is this config option meant to use dashes or underscores to separate words?

richardwhiuk:
Don't you need a change to `debian/homestead.init.d` to pass the config option in?

eleanor-merry:
D'oh - I ran git add in homestead/src by mistake. I've added the init.d and include changes now

src/main.cpp
eleanor-merry: The config option that's passed into Homestead uses dashes, and the config option in the shared config file uses underscores. The init.d file (which I'd forgotten to check in) reads the shared config and sets the Homestead configuration as appropriate

KristaHyer: Oh cool that makes sense - thanks!

debian/homestead.init.d
```diff
@@ -160,6 +160,7 @@ get_daemon_args()
 
 
         [ "$sas_use_signaling_interface" != "Y" ] || sas_signaling_if_arg="--sas-use-signaling-interface"
+        [ "$support_shared_ifc" == "N" ] || support_shared_ifcs_arg="--support-shared-ifcs"
```
richardwhiuk: Should we do != Y to match the `sas_user_signaling_interface` here? I think it's probably better if 'n' and 'no' don't match, rather than 'yes' matching?

Alternatively, should we have some common code which checks whether a value is truthy?

PR #452, opened 2017-05-26
[Reviewer: Adam] Cassandra fixes

sebrexmetaswitch:
There are 2 parts:

### Don't send 503 HTTP responses on Cassandra connection errors (first 2 commits)
In split-storage, we retry to multiple cassandra nodes already, so we don't want Sprout to retry if there's an error.
I tested this by creating the error conditions manually, and making calls, verifying that we do indeed get 504 responses sent.

### Use the new HAOperation cassandra operation (3rd commit)
This is the corresponding change to https://github.com/Metaswitch/cpp-common/pull/649
Testing is described in that PR.

### DIFF
--- a/include/cache.h
+++ b/include/cache.h
@@ -199,7 +199,7 @@ public:
     return new PutAuthVector(private_id, auth_vector, timestamp, ttl);
   }
 
-  class GetRegData : public CassandraStore::Operation
+  class GetRegData : public CassandraStore::HAOperation
   {
   public:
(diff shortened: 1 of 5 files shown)


johadalin:
:+1: LGTM

sebrexmetaswitch:
Just the UTs I've added to this, to go with the latest commit in https://github.com/Metaswitch/cpp-common/pull/649

johadalin:
UTs look decent. think they cover the new changes well :shipit:

PR #456, opened 2017-06-08
[Reviewer: Mat] Update PD logs for split-storage

sebrexmetaswitch:


include/homestead_pd_definitions.h
```diff
@@ -74,14 +74,13 @@ static const PDLog1<int> CL_HOMESTEAD_CASSANDRA_CACHE_INIT_FAIL
 (
   PDLogBase::CL_HOMESTEAD_ID + 6,
   LOG_ERR,
-  "Fatal - Failed to initialize the cache for the CassandraStore - "
+  "Fatal - Failed to initialize the Cassandra cache - "
```
MatMeredith: I had to go and look at the code to figure out what this log is actually reporting.   It looks like the most likely reason that it will be raised is failure to connect to the Cassandra service, rather than anything to do with "initializing a Cache".  

Looks like you've rightly changed the recommended actions to reflect that the most likely issue here is now a connectivity one (where it wasn't before) -- suggest we should update the description of the issue similarly?

sebrexmetaswitch: I've tweaked it to refer to the cassandra "connection", which is a more accurate description of what the `Cache` in homestead is.

PR #461, opened 2017-07-14
PPR should update with only a charging address

LucyBarcroft:
PPR was failing to update the charging address when there was no IMS subscription element present. It should be able to update without this. Was previously failing to find any associated public IDs, since this function is only used when subscribers are locally provisioned. Changed such that it can correctly find the default IDs, then uses the Default ID to identify the Implicit Registration Set in order to update the charging address, rather than obtaining a list of all IDs. 

Tested for a simple subscription, with only one IMPI and one IMPU and it updates the charging information correctly. It also appears to update correctly with an IMPI and one implicit registration set, containing multiple IMPUs, although further testing is required.

### DIFF
--- a/src/handlers.cpp
+++ b/src/handlers.cpp
@@ -2322,7 +2322,7 @@ void PushProfileTask::run()
   SAS::report_event(ppr_received);
 
   // Received a Push Profile Request. We may need to update an IMS
-  // subscription and/or charging address information in the cache.
+  // subscription and/or charging address information in the cache
   _ims_sub_present = _ppr.user_data(_ims_subscription);
   _charging_addrs_present = _ppr.charging_addrs(_charging_addrs);
 
@@ -2339,7 +2339,7 @@ void PushProfileTask::run()
   // will change the default public id. If it will, reject it, otherwise
   // continue.
   _impi = _ppr.impi();
- 
+
   if ((!_charging_addrs_present) && (!_ims_sub_present))
   {
     send_ppa(DIAMETER_REQ_SUCCESS);
@@ -2391,11 +2391,10 @@ void PushProfileTask::update_reg_data()
   // Get the default public id from the ims subscription if it is present.
   // If the ims subscription is not present, take the first of the default ifcs
   // found from searching each IRS associated with the IMPI on the PPR.
-  // Currently taking the list of IMPUs when there is no IMS sub present to be 
+  // Currently taking the list of IMPUs when there is no IMS sub present to be
   // the default ID, since this can be used to identify the set and the IMS
   // sub does not change.
-  // TODO - THIS IS INCORRECT AS THE CORRECT DEFAULT ID MUST BE FOUND.
-  // CURRENTLY THIS CANNOT BE FIXED AS THE CODE IS BROKEN.
+
   std::string default_public_id;
   if (_ims_sub_present)
   {
@@ -2404,9 +2403,9 @@ void PushProfileTask::update_reg_data()
   else
   {
     default_public_id = _default_impus.front();
-    _impus.push_back(default_public_id);  
+    _impus.push_back(default_public_id);
   }
- 
+
   Cache::PutRegData* put_reg_data =
     _cfg->cache->create_PutRegData(_impus,
                                    default_public_id,
@@ -2418,12 +2417,12 @@ void PushProfileTask::update_reg_data()
   event.add_var_param(impus_str);
 
   if (_ims_sub_present)
-  {  
+  {
    TRC_INFO("Updating IMS subscription from PPR");
    put_reg_data->with_xml(_ims_subscription);
    event.add_compressed_param(_ims_subscription, &SASEvent::PROFILE_SERVICE_PROFILE);
   }
-  else
+   else
   {
     event.add_compressed_param("IMS subscription unchanged", &SASEvent::PROFILE_SERVICE_PROFILE);
   }
@@ -2437,9 +2436,24 @@ void PushProfileTask::update_reg_data()
     event.add_var_param(_charging_addrs.log_string());
     put_reg_data->with_charging_addrs(_charging_addrs);
   }
+
   else
   {
     event.add_var_param("Charging addresses unchanged");
+  }
+
+  // if there are multiple implicit registration sets, create a new vector, containing any further default IMPUs, that have yet
+  // to be updated with the charging addresses.
+  if (_charging_addrs_present)
+  {
+     for (std::vector<std::string>::iterator it = _default_impus.begin(); it != _default_impus.end(); it++)
+     {
+       if (*it != default_public_id)
+       {
+     	 _further_impus.push_back(std::string(*it));
+	 std::string test = _further_impus.back();
+       }
+     }
   }
 
   CassandraStore::Transaction* tsx =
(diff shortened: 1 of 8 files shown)


src/handlers.cpp
```diff
@@ -2366,136 +2357,70 @@ void PushProfileTask::run()
...
-                                           CassandraStore::ResultCode error,
-                                           std::string& text)
-{
-  SAS::Event event(this->trail(), SASEvent::CACHE_GET_ASSOC_IMPU_FAIL, 0);
-  SAS::report_event(event);
-  TRC_DEBUG("Cache query failed with rc %d", error);
-  send_ppa(DIAMETER_REQ_FAILURE);
-}
 
 void PushProfileTask::update_reg_data()
```
eleanor-merry: This comment hasn't been addressed

src/handlers.cpp
```diff
@@ -2491,11 +2423,13 @@ void PushProfileTask::update_reg_data()
     event.add_var_param(_charging_addrs.log_string());
     put_reg_data->with_charging_addrs(_charging_addrs);
   }
+
```
eleanor-merry: No blank line here.

src/handlers.cpp
```diff
@@ -2491,11 +2423,13 @@ void PushProfileTask::update_reg_data()
...
   else
   {
     event.add_var_param("Charging addresses unchanged");
   }
 
+
```
eleanor-merry: No blank line here

src/handlers.cpp
```diff
@@ -2506,6 +2440,63 @@ void PushProfileTask::update_reg_data()
   SAS::report_event(event);
 }
 
+void PushProfileTask::get_registration_set()
```
eleanor-merry: Can you use irs or implicit_registration_set consistently in all your variable/function names?

src/handlers.cpp
```diff
@@ -2518,6 +2497,7 @@ void PushProfileTask::on_get_primary_impus_failure(CassandraStore::Operation* op
 
 void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op)
 {
+  // will return a list of default IMPUs, one default IMPU per IRS.
```
eleanor-merry: This comment hasn't been addressed.

src/handlers.cpp
```diff
@@ -2529,59 +2522,171 @@ void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op
...
   {
-    // Attempt to find a match between any of the default ids of the sets the
-    // private id corresponds to, and the default id in the PPR. If no match is
-    // found the PPR is changing the default id (which is not permitted), so
-    // reject it. Otherwise, continue.
-    std::string new_default_id;
-    XmlUtils::get_default_id(_ims_subscription, new_default_id);
-    bool impu_match_found = false;
-    for (std::string current_default_id : _default_impus)
+    if (default_ids_match())
```
eleanor-merry: impus not ids surely? In general your functions names/variables use ids where I think they should use impus for accuracy.

src/handlers.cpp
```diff
@@ -2529,59 +2522,171 @@ void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op
...
+  {
+    no_ims_set_first_default();
+    get_registration_set();
+  }
+}
 
-    if (!impu_match_found)
+bool PushProfileTask::default_ids_match()
+{
+  // Attempt to find a match between any of the default ids of the sets the
```
eleanor-merry: This comment is in the wrong place - this function is only responsible for checking if there's a match, it's the place that calls this function that then rejects the PPR.

src/handlers.cpp
```diff
@@ -2529,59 +2522,171 @@ void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op
...
+
+  if (!found_sip_uri)
+  {
+    TRC_ERROR("No SIP URI in Implicit Registration Set");
+    SAS::Event event(this->trail(), SASEvent::NO_SIP_URI_IN_IRS, 0);
+    event.add_compressed_param(_ims_subscription, &SASEvent::PROFILE_SERVICE_PROFILE);
+    SAS::report_event(event);
+  }
+
+  XmlUtils::get_default_id(_ims_subscription, _default_public_id);
```
eleanor-merry: You should comment here why you're removing the first default IMPU

src/handlers.cpp
```diff
@@ -2529,59 +2522,171 @@ void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op
...
+  _first_default_id = _default_public_id;
+  _default_impus.erase(std::remove(_default_impus.begin(),
+                                   _default_impus.end(),
+                                   _default_public_id),
+                       _default_impus.end());
+  _default_impus.push_back(_first_default_id);
+}
+
+void PushProfileTask::no_ims_set_first_default()
```
eleanor-merry: I'm not sure this warrants being in its own function - I think the code flow would be easier to understand if it was just these two lines called directly (with an appropriate comment).

src/handlers.cpp
```diff
@@ -2529,59 +2522,171 @@ void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op
...
   SAS::Event event(this->trail(), SASEvent::UPDATED_REG_DATA, 0);
   SAS::report_event(event);
-  send_ppa(DIAMETER_REQ_SUCCESS);
+  if (should_send_ppa())
+  {
+    send_ppa(DIAMETER_REQ_SUCCESS);
+  }
+
+  if (_charging_addrs_present)
+  {
```
eleanor-merry: You could just call get_registration_set - let that function decide what action to take rather than repeat the logic here.

src/ut/handlers_test.cpp
```diff
@@ -75,6 +75,8 @@ class HandlersTest : public testing::Test
   static const std::string IMPI;
   static const std::string IMPU;
   static std::vector<std::string> IMPU_IN_VECTOR;
+  static std::vector<std::string> IMPU2_IN_VECTOR;
```
eleanor-merry: Nice! - I think the changes you've made here really make the UTs easier to follow.

src/handlers.cpp
```diff
@@ -2529,59 +2522,171 @@ void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op
...
+void PushProfileTask::delete_impus()
+{
+  CassandraStore::Operation* delete_IMPU =
+    _cfg->cache->create_DeleteIMPUs(_impus_to_delete,
+			       Cache::generate_timestamp());
+  CassandraStore::Transaction* tsx = new CacheTransaction;
+  _cfg->cache->do_async(delete_IMPU, tsx);
+}
+
+bool PushProfileTask::check_impus_added()
```
eleanor-merry: Is check the right word? How about 'any_new_impus'?

src/handlers.cpp
```diff
@@ -2529,59 +2522,171 @@ void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op
...
+  {
+    for (std::vector<std::string>::iterator sub_impu = _impus.begin();
+         sub_impu != _impus.end();
+         sub_impu++)
+    {
+      if (*irs_impu == *sub_impu)
+      {
+        break;
+      }
+      if (sub_impu+1 == _impus.end())
```
eleanor-merry: Blank line between the two ifs please

src/handlers.cpp
```diff
@@ -2529,59 +2522,171 @@ void PushProfileTask::on_get_primary_impus_success(CassandraStore::Operation* op
...
+      break;
     }
   }
+  return impu_match_found;
+}
 
-  // If we've reached this point, we can safely update the reg data in the
-  // knowledge it will not change the default public identity.
-  update_reg_data();
+void PushProfileTask::no_default_id_match_send_failure()
```
eleanor-merry: I'm not sure this warrants being in its own function (it's hard to come up with a good name for it!). Maybe just have these lines in place?

PR #465, opened 2017-08-01
[Reviewer: SR2, RJW2] Cassandra to Memcached

richardwhiuk:
_Edit (Seb Rex): Added a description to this PR for posterity_

This PR changes Homestead to use Memcached rather than Cassandra for caching subscriber data.

Previously, when using Homestead-Prov, the data was pre-populated in the cache so the Handlers could skip the HSS lookup and read directly from the cache.

Now, the Homestead-Prov data is still set in Cassandra. Homestead now treats Homestead-Prov like an HSS, reading the data from Cassandra and caching it in Memcached, just as it would for a real HSS.

Two new interfaces are added, which are used by the Handlers:
* `HssConnection` - this is the interface that represents a connection to an HSS, There are 2 implementations:
    * `DiameterHssConnection` for connections to a real HSS
    * `HsProvHssConnection` for connections to the Cassandra database used for HSProv
* `HssCache` - this is the interface that represents a connection to the cache of subscriber data. There is one implementation (at the moment):
    * `MemcachedCache` for using Astaire/Memcached to cache the subscriber data


include/hss_cache.h
```diff
@@ -0,0 +1,76 @@
...
+  virtual ~HssCache();
+
+  // All of these methods are synchronous, and run on a thread that is OK to block
+  // They return Store::Status (from cpp-common's Store) which is used to determine which callback to use
+  // If they are getting/listing data, the data is put into the supplied datastructure
+
+  // Give a set of public IDs (representing an implicit registration set) an associated private ID.
+  // used for regdata task
+  // only adds public id to impi_mapping table
+  Store::Status put_associated_private_id(std::vector<std::string> impus, std::string default_public_id, std::string impi, int ttl);
```
richardwhiuk: Are these always default IMPUs?

sebrexmetaswitch:
I've updated the HssCacheProcessor to represent the interface we discussed, where homestead will simply get objects from the cache, modify them, and write them back.

Can you take a look at hss_cache_processor.h and the ims_subscription.h and implicit_reg_set.h which go with it? The HssCache hasn't been updated, buts its functions will look basically the same as the processor's ones without the callbacks.

include/hss_cache_processor.h
```diff
@@ -0,0 +1,106 @@
...
+  // Lists impus, starting at starting_from, limited to count
+  void list_impus(std::function<void(std::vector<std::string>*)> success_cb,
+                  std::function<void(Store::Status)> failure_cb,
+                  int count,
+                  int starting_from);
+private:
+  // Dummy exception handler callback for the thread pool
+  static void inline exception_callback(std::function<void()> callable)
+  {
+  }
```
richardwhiuk: Surely this should log or something?

include/hss_connection.h
```diff
@@ -0,0 +1,131 @@
...
+  // Send a multimedia auth request to the HSS
+  virtual void send_multimedia_auth_request(std::function<void(MultimediaAuthResponse*)> callback,
+                                            std::string impi,
+                                            std::string impu,
+                                            std::string provided_server_name,
+                                            std::string scheme,
+                                            std::string authorization);
+
+  // Send a user authorization request to the HSS
+
```
richardwhiuk: random blank line.

include/hss_connection.h
```diff
@@ -0,0 +1,131 @@
...
+  // Each method takes a callback, which will be called on a different thread
+  // (in general) once the request is complete.
+  // Ownership of the response objects passed to the callbacks is delegated to
+  // the callback target. It's the responsibility of the callback to delete the
+  // object when it is done with it.
+  // ---------------------------------------------------------------------------
+
+  // Send a multimedia auth request to the HSS
+  virtual void send_multimedia_auth_request(std::function<void(MultimediaAuthResponse*)> callback,
+                                            std::string impi,
```
richardwhiuk: Should we wrap the variables up in a struct - MultimediaAuthRequest for example?

include/hss_connection.h
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+private:
+  HssResultCode _result_code;
+};
+
+// Subclasses of the HssResponse.
+// We will have 4 - one for each of the 4 requests we make to the HSS.
+// Each one holds the data that the handlers need to process the response.
+class MultimediaAuthResponse : public HssResponse
```
richardwhiuk: This is confusing - MAR may mean MultimediaAuthRequest (in Diameter) or MultimediaAuthResponse (in this class). The Diameter equivalent is MultimediaAuthAnswer.

I think we either need different names which describe what they actually do (eg. GetAuthDetails, GetProfile, LookupAuthDetails, FindUser etc), or we need to follow the Diameter names exactly.

sebrexmetaswitch:
*[src/memcached_cache.cpp, line 163 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkUSw_0DyPLDUxoGPu:-KrkUSw_0DyPLDUxoGPv:b-xzucjv) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L163)):*
> ```cpp
>                       _associated_impus,
>                       state);
> }
> ```

Are these actually what we want to be doing?
If we have updated the TTL, then our data is newer, but this will override our stored data with "DELETED"

e.g.
* We had impi1 before, and now have both impi1 and impi2 which we're trying to store
* before we enter this function, our `_impis` Data contains

```
impi1: UNCHANGED,
impi2: ADDED
```

* when we enter this function, we note that the TTL has been updated and so we set `state` to `DELETED`
* when we load read the impi data from the store's Impu, we'll update our `_impi` Data to have:

```
impi1: DELETED,
impi2: ADDED
```
which isn't right

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/homestead/465)*

sebrexmetaswitch:
Reviewed 4 of 61 files at r1, 1 of 7 files at r5.
Review status: 4 of 61 files reviewed at latest revision, 22 unresolved discussions, some commit checks failed.

---

*[include/memcached_cache.h, line 21 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrjdyX76JVN2d_VKwYY:-KrjdyX8qT_0Q1rZeuia:b49ybum) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L21)):*
> ```objc
> #include <vector>
> 
> class MemcachedImsSubscription : public ImsSubscription
> ```

This class needs to implement to virtual methods `void set_charging_addrs(ChargingAddresses new_addresses)` and `ImplicitRegistrationSet* get_irs_for_default_impu(std::string impu)`, as those are the 2 methods that the PPR task will use

---

*[include/memcached_cache.h, line 24 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krjf5_dHU7FHC3VZeux:-Krjf5_dHU7FHC3VZeuy:b-qd4exg) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L24)):*
> ```objc
> {
> public:
>   MemcachedImsSubscription(ImpuStore::ImpiMapping*& mapping,
> ```

Do we need to be passing a pointer by reference?

---

*[include/memcached_cache.h, line 87 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krjh048VrPJsMacSMGn:-Krjh048VrPJsMacSMGo:b-lguaby) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L87)):*
> ```objc
>     for (const std::pair<std::string, State>& entry : _impis)
>     {
>       if (entry.second == State::UNCHANGED || entry.second == State::DELETED)
> ```

As discussed, should be ADDED not DELETED

---

*[include/memcached_cache.h, line 102 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krjh9T86I9C8MY1udiX:-Krjh9T86I9C8MY1udiY:b-8a05gs) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L102)):*
> ```objc
>     for (const std::pair<std::string, State>& entry : _associated_impus)
>     {
>       if (entry.second == State::UNCHANGED || entry.second == State::DELETED)
> ```

As above

---

*[include/memcached_cache.h, line 162 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkQFUdjmUpdac5dSqQ:-KrkQFUdjmUpdac5dSqR:b-ddh84t) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L162)):*
> ```objc
>   ImpuStore::DefaultImpu* get_impu_from_impu(ImpuStore::Impu* with_cas);
> 
>   // Get an IMPU for this IRS representing the given store
> ```

Maybe update this comment to say that it returns an Impu if the IRS was created from the given store (and hence the CAS value is valid) or `nullptr` if not (ie. if was created for a different store)

---

*[include/memcached_cache.h, line 181 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkRTLibdna1Z3Feqgb:-KrkRTLibdna1Z3Feqgc:b8499c2) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L181)):*
> ```objc
>   };
> 
>   typedef std::map<std::string, State> Data;
> ```

It would be good to get a comment explaining what data this represents.

---

*[include/memcached_cache.h, line 187 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkN-mPfRcx66gS3Kw5:-KrkN-mPfRcx66gS3Kw6:b-3p86tc) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L187)):*
> ```objc
>   const uint64_t _cas;
> 
>   static std::vector<std::string> get(const Data& data, State status)
> ```

This function is just called `get`, which isn't hugely descriptive. A comment/better name would be nice :)

---

*[src/memcached_cache.cpp, line 65 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkQyNrdEvKl4wlnDlf:-KrkQyNrdEvKl4wlnDlg:bkxkg6m) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/src/memcached_cache.cpp#L65)):*
> ```cpp
>                        MemcachedImplicitRegistrationSet::Data data,
>                        MemcachedImplicitRegistrationSet::State state)
> {
> ```

This is (necessarily) quite generic, so can we have a comment explaining what it does? I think that `store` will always be a vector of IDs (impis or impus), so renaming that to `store_ids` and `state` to `default_state` might be helpful

---

*[src/memcached_cache.cpp, line 81 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkS-QYgv7Ufa1_fZ4u:-KrkS-QYgv7Ufa1_fZ4v:ba5kzi8) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L81)):*
> ```cpp
> }
> 
> void set_elements(const std::vector<std::string>& updated,
> ```

Again, a comment. I think I've just figured it out though, so something like:

```
// Updates data with the new elements specified in "updated"
// Elements present in both are left with State `UNCHANGED`
// Elements present in `updated` but not in `data` are set to State `ADDED`
// Elements present in `data` but not in `updated` are set to State `DELETED`
```

---

*[src/memcached_cache.cpp, line 117 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkT6J1Wfn2G-gYlZMs:-KrkT6J1Wfn2G-gYlZMt:b-tv3jz1) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L117)):*
> ```cpp
> }
> 
> void MemcachedImplicitRegistrationSet::update_from_store(ImpuStore::DefaultImpu* impu)
> ```

Maybe consider renaming this to `update_from_stored_impu`?

---

*[src/memcached_cache.cpp, line 139 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkTCC20ZRJSQNpboUX:-KrkTCC20ZRJSQNpboUY:b-mqtbeq) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L139)):*
> ```cpp
>   {
>     // If we are marked as refreshed, then the data from the store is *less* up
>     // to date than our data, so we should mark it as changed.
> ```

Nice comment :)
Can we get something similar a few lines above, when we're deciding whether to update our member variables with the data from `impu`?

---

*[src/memcached_cache.cpp, line 246 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkXBBuizYWDYCKG0Go:-KrkXBBuizYWDYCKG0Gp:b-ljzcjk) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L246)):*
> ```cpp
>     {
>       bool is_default = data->is_default_impu();
>       bool is_associated = (is_default && ((ImpuStore::DefaultImpu*)data)->has_associated_impu(impu));
> ```

I can see what these are doing, but a comment explaining why would be nice

---

*[src/memcached_cache.cpp, line 284 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkYcqwJTJuHsP8Jm8V:-KrkYcqwJTJuHsP8Jm8W:b-crhwxz) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L284)):*
> ```cpp
>   {
>     ImpuStore::ImpiMapping* mapping = get_impi_mapping_gr(impi, trail);
>     get_implicit_registration_sets_for_impus(mapping->get_default_impus(), trail, result);
> ```

Can we be smarter here?
Two impis could have registered the same default impu, so is it worth getting all of the default impus first, and then getting the IRSs?

---

*[src/memcached_cache.cpp, line 298 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkZ7kV5yHpCGp0t-BA:-KrkZ7kV5yHpCGp0t-BB:b-sof350) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L298)):*
> ```cpp
>     ImplicitRegistrationSet* irs;
>     get_implicit_registration_set_for_impu(impu, trail, irs);
>     result.push_back(irs);
> ```

We probably want to spot duplicates here too. At the least, we could check whether we've already pushed an IRS for a given default impu before pushing the next.

Imagine if we're asked to get IRSs for `{impu2, impu1}`  and `impu2` is actually just an associated impu of `impu1` It would be nice if we could:
* lookup `impu2` in the cache
* notice it's a Assoc. impu, and so lookup it's default impu (`impu1`) in the cache (which we currently do)
* when we come to `impu1` in the vector, notice that we've already done this lookup and skip it

---

*[src/memcached_cache.cpp, line 343 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krkb0SwU0xor6WpN0Hm:-Krkb0SwU0xor6WpN0Hn:b-3ao4nt) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L343)):*
> ```cpp
>   Store::Status status = Store::Status::OK;
> 
>   // Remove old IMPI mappings
> ```

Can we also put in a comment here explaining why all these impi mapping changes are using CASed writes, as discussed?

---

*[src/memcached_cache.cpp, line 389 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krkcm8OniOg_SMi_bk-:-Krkcm8OniOg_SMi_bk0:b-gorfgu) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L389)):*
> ```cpp
>           if (!mapping->has_default_impu(irs->default_impu))
>           {
>             mapping->add_default_impu(irs->default_impu);
> ```

If these are the unchanged impis, how would the default impu not already be in the mapping?

---

*[src/memcached_cache.cpp, line 409 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkfW3DXvhBoSkR_5_2:-KrkfW3EgIVySKfsMz6M:b6bdqut) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L409)):*
> ```cpp
>   {
>     int64_t expiry = time(0) + irs->get_ttl();
>     ImpuStore::ImpiMapping* mapping = new ImpuStore::ImpiMapping(impi,
> ```

It feels like you've made a decision that it's more likely that we're writing a new IMPI mapping entry than adding to an existing one, because we're attempting to write first, and then only reading if that was contended.
Fine if so, but a comment explaining that it was deliberate would be good :)

---

*[src/memcached_cache.cpp, line 451 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkiJ9ql-VnxfUssa4L:-KrkiJ9ql-VnxfUssa4M:b-u5vh6v) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L451)):*
> ```cpp
>     do
>     {
>       ImpuStore::Impu* mapping = store->get_impu(associated_impu, trail);
> ```

It's a bit sad that we're having to do a read first, then an un-CASed delete

As discussed, this read first is a defence against the possibility of this Assoc Impu having been associated to another IRS between us getting the data from the cache and setting the new data here (which may have an SAR/SAA in between, so could be quite long)

However, with the code as it is we've only reduced the window, not closed it, since the data could (in theory) change between us reading it and deleting it.

This scenario feels unlikely to me, so I wonder whether we should just be doing a blind delete without reading first. If we think that we want to defend against this scenario, then i think we should be doing a CASed delete instead of un-CASed, as that would close the window entirely

---

*[src/memcached_cache.cpp, line 532 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krkkmx-mWoewhn_hZCW:-Krkkmx-mWoewhn_hZCX:b-6x120a) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L532)):*
> ```cpp
>     else
>     {
>       // The IRS was not originally created from this store, so we don't
> ```

"...or we had a conflict"

---

*[src/memcached_cache.cpp, line 539 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krkle5UACiN1i7wlBWV:-Krkle5UACiN1i7wlBWW:byc4thu) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L539)):*
> ```cpp
>       ImpuStore::Impu* mapped_impu = store->get_impu(irs->default_impu, trail);
> 
>       if (mapped_impu->is_default_impu())
> ```

`get_impu` can return `nullptr` if there's nothing in the store, which I think will happen if we're writing to a remote store for the first time. So presumably we have to protect against this by checking for `nullptr` and creating a new one if so

---

*[src/memcached_cache.cpp, line 594 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkmMVOojc0Ir-CL3C-:-KrkmMVOojc0Ir-CL3C0:bivia71) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L594)):*
> ```cpp
> }
> 
> Store::Status MemcachedCache::put_implicit_registration_set(MemcachedImplicitRegistrationSet* irs,
> ```

May be worth a comment that these three operations happen in sequence, and we can't guarantee consistency between the three tables

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/homestead/465)*

sebrexmetaswitch:
*[include/memcached_cache.h, line 64 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkuH5QnuxHNxihorVN:-KrkuH5QnuxHNxihorVO:b-mdv9ef) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/include/memcached_cache.h#L64)):*
> ```objc
> {
> public:
>   MemcachedImplicitRegistrationSet(ImpuStore::DefaultImpu* default_impu) :
> ```

Also, we need to be setting the XML, Charging Addrs and reigstration state when we create it from an Impu we retrieved from the store

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/homestead/465)*

sebrexmetaswitch:
Reviewed 2 of 3 files at r6.
Review status: 6 of 61 files reviewed at latest revision, 34 unresolved discussions, some commit checks failed.

---

*[src/impu_store.cpp, line 48 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl-ontH_rC5v-_eu6a:-Krl-ontH_rC5v-_eu6b:bevmncv) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L48)):*
> ```cpp
> 
> // The maximum buffer size to use for IMPU compression
> static const int MAX_BUFFER_LEN = 131072;
> ```

Did this number come from anywhere in particular?

---

*[src/impu_store.cpp, line 50 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl0ImiJz31G9vTxG7K:-Krl0ImiJz31G9vTxG7L:bzdk0vm) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L50)):*
> ```cpp
> static const int MAX_BUFFER_LEN = 131072;
> 
> void encode_varbyte(uint64_t uncomp_size, std::string& data)
> ```

A comment on this and the decode counterpart explaining why they do this would be good

---

*[src/impu_store.cpp, line 121 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl1bKJne_l4VSjPfE2:-Krl1bKJne_l4VSjPfE3:b98tajx) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L121)):*
> ```cpp
> 
>   // Version is stored in character 0.
>   if (data[0] == 0)
> ```

is this comparison always safe? (We're comparing a `char` to an `int`, right? We expect it to be small, so probably aren't too worried about `signed` vs `unsigned` `chars`, but wanted to check)

---

*[src/impu_store.cpp, line 258 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl3Btoi_Bqr7Qmg0GD:-Krl3Btoi_Bqr7Qmg0GE:b-l0brnx) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L258)):*
> ```cpp
>   JSON_SAFE_GET_STRING_MEMBER(json, JSON_SERVICE_PROFILE, service_profile);
> 
>   extract_array(json, JSON_ASSOCIATED_IMPUS, assoc_impus);
> ```

This has 2 `JSON_ASSERT_*`s, which will throw exceptions I think. So we should probably have a `try...catch...` here

---

*[src/impu_store.cpp, line 320 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl4CJ02y0F3-mdH9N2:-Krl4CJ02y0F3-mdH9N3:b-rqbo95) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L320)):*
> ```cpp
>                     uncomp_size, buffer_length,
>                     new_buffer_length, MAX_BUFFER_LEN);
>         break;
> ```

When we `break` here, don't we need to do `LZ4_resetStream(stream)` too to make sure it's in a good state next time? Or is it fine to leave it not reset?

---

*[src/impu_store.cpp, line 374 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl4oNqWhUxb-h7uW7a:-Krl4oNqWhUxb-h7uW7b:b7p44s5) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L374)):*
> ```cpp
>   // LCOV_EXCL_START
>   // This only happens when we fail to compress some data,
>   // which isn't hittable in the UTs
> ```

can we force the UTs to have a smaller max buffer size or something to hit this? Or do you think it's not interesting enough to test?

---

*[src/impu_store.cpp, line 398 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl5LcpyKBYXpIohR8k:-Krl5LcpyKBYXpIohR8l:bcmspjc) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L398)):*
> ```cpp
>   int uncomp_size_len = (uncomp_size_bits + 6) / 7;
> 
>   data.reserve(1 + uncomp_size_len + uncomp_size);
> ```

Shouldn't we be using `comp_size` here? Since we know exactly how much space we'll need?

---

*[src/impu_store.cpp, line 472 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl6S0Kb2fwOiMP9EFe:-Krl6S0Kb2fwOiMP9EFf:bb15dxo) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L472)):*
> ```cpp
> }
> 
> ImpuStore::ImpiMapping* ImpuStore::ImpiMapping::from_data(std::string const& impi,
> ```

Worth a comment pointing out that this one is uncompressed, which is why it's just a JSON string

---

*[src/impu_store.cpp, line 498 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl6SruMQr2OGR9C4K2:-Krl6SruMQr2OGR9C4K3:b2o3adb) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L498)):*
> ```cpp
> }
> 
> Store::Status ImpuStore::ImpiMapping::to_data(std::string& data)
> ```

as previous comment

---

*[src/impu_store.cpp, line 643 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl7_sJJN2CnyRLFIqk:-Krl7_sJJN2CnyRLFIql:ba9339c) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L643)):*
> ```cpp
>   int64_t expiry = 0L;
> 
>   JSON_ASSERT_ARRAY(json[JSON_DEFAULT_IMPUS]);
> ```

Again, I think we need to `try...catch...` this. But can't you use your `extract_array()` helper function you defined above?

---

*[src/impu_store.cpp, line 664 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl8TnbxU5JmO4G7zAJ:-Krl8TnbxU5JmO4G7zAK:bpp5wg7) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L664)):*
> ```cpp
> void ImpuStore::ImpiMapping::write_json(rapidjson::Writer<rapidjson::StringBuffer>& writer)
> {
>   writer.String(JSON_DEFAULT_IMPUS);
> ```

Again, may as well use `write_array` from above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/homestead/465)*

richardwhiuk:
Review status: 4 of 61 files reviewed at latest revision, 34 unresolved discussions, some commit checks failed.

---

*[src/impu_store.cpp, line 121 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl1bKJne_l4VSjPfE2:-KrlEoz2-hpZiTp96DDe:b5njz10) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L121)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

is this comparison always safe? (We're comparing a `char` to an `int`, right? We expect it to be small, so probably aren't too worried about `signed` vs `unsigned` `chars`, but wanted to check)
</blockquote></details>

Yes, it should always be safe. The char will get promoted, but that's fine.

---

*[src/impu_store.cpp, line 258 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl3Btoi_Bqr7Qmg0GD:-KrmAhsa8MfRcC9IdQSW:b-vbt1z0) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L258)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

This has 2 `JSON_ASSERT_*`s, which will throw exceptions I think. So we should probably have a `try...catch...` here
</blockquote></details>

I've checked with an if block in `extract_array`, rather than catching the exception.

---

*[src/impu_store.cpp, line 320 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl4CJ02y0F3-mdH9N2:-KrmB-bReMDAWlJryxGY:b-o005rm) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L320)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

When we `break` here, don't we need to do `LZ4_resetStream(stream)` too to make sure it's in a good state next time? Or is it fine to leave it not reset?
</blockquote></details>

No - the next thing we do with this stream is free it. We create a new stream (based on `_thrd_lz4_stream` each time.

We only need to reset it if we are going to reuse it (e.g. because we have expanded the buffer).

---

*[src/impu_store.cpp, line 374 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl4oNqWhUxb-h7uW7a:-KrmBOUdpt_K76h1sAnN:b1u91hm) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L374)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

can we force the UTs to have a smaller max buffer size or something to hit this? Or do you think it's not interesting enough to test?
</blockquote></details>

I don't think it's interesting enough to test - the only failure is if lz4 fails to compress some data. We could change MAX_BUFFER_LENGTH, but it's quite a lot of pain for a really difficult to see edge condition.

---

*[src/impu_store.cpp, line 398 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl5LcpyKBYXpIohR8k:-KrmBp36PTndAvQNDOPR:b-dho2z6) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L398)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Shouldn't we be using `comp_size` here? Since we know exactly how much space we'll need?
</blockquote></details>

Yep - that'll make the buffer too big, and potentially cause a resize if compressing makes life worse (which is unlikely, but possible).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/homestead/465)*

richardwhiuk:
Review status: 4 of 61 files reviewed at latest revision, 34 unresolved discussions, some commit checks failed.

---

*[include/memcached_cache.h, line 21 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrjdyX76JVN2d_VKwYY:-KrmEoHKQyNvVaEYVb85:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L21)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

This class needs to implement to virtual methods `void set_charging_addrs(ChargingAddresses new_addresses)` and `ImplicitRegistrationSet* get_irs_for_default_impu(std::string impu)`, as those are the 2 methods that the PPR task will use
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 163 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkUSw_0DyPLDUxoGPu:-KrmKD1uYD8bu76Ad6cr:b4souh6) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L163)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Are these actually what we want to be doing?
If we have updated the TTL, then our data is newer, but this will override our stored data with "DELETED"

e.g.
* We had impi1 before, and now have both impi1 and impi2 which we're trying to store
* before we enter this function, our `_impis` Data contains

```
impi1: UNCHANGED,
impi2: ADDED
```

* when we enter this function, we note that the TTL has been updated and so we set `state` to `DELETED`
* when we load read the impi data from the store's Impu, we'll update our `_impi` Data to have:

```
impi1: DELETED,
impi2: ADDED
```
which isn't right
</blockquote></details>

`update_from_store` uses `emplace` which only changes an element if it doesn't exit.

However, there is an issue, which is that deletes might not persist if a HS performs a deregistration while a different HS performs a registration.

 - HS 1 -> GET IMPU - returns {1}
 - HS 2 -> GET IMPU - returns {1}
 - HS 2 -> DEREGISTRATION of 1
 - HS 2- > WRITE IMPU {} - succeeds
 - HS 1 -> REGISTRATION of 2 {1U, 2A }
 - HS 1 -> WRITE IMPU {1, 2} - fails consistency check
 - HS 1 -> READ IMPU {} - {1U, 2A} - CAS updated
 - HS 1 -> WRITE IMPU {1, 2} - 2 just got unregistered

The problems is that a registration isn't prima facie valid for IMPIS - Homestead doesn't authoritatively know a valid state which the store can put stuff in.

I think the Handlers need to say 'MARK THIS IMPI as ADDED', and 'MARK THIS IMPI as DELETED', rather than 'GET IMPIS' and 'SET IMPIS'.

---

*[src/memcached_cache.cpp, line 389 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krkcm8OniOg_SMi_bk-:-KrmPM4Q_97Rd9Kh1Vi7:b-8puzsf) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L389)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

If these are the unchanged impis, how would the default impu not already be in the mapping?
</blockquote></details>

They should be, but we should aim to leave the data store in a consistent state, and we need to do a write anyway to update the expiry value, which thus has to be CASed.

---

*[src/memcached_cache.cpp, line 409 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkfW3DXvhBoSkR_5_2:-KrmQbfQZspqsOpYZM1Q:b-42vxdg) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L409)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

It feels like you've made a decision that it's more likely that we're writing a new IMPI mapping entry than adding to an existing one, because we're attempting to write first, and then only reading if that was contended.
Fine if so, but a comment explaining that it was deliberate would be good :)
</blockquote></details>

Shrug. I think multiple IMPI-IMPU mappings are rare, and thus given we think it's new, this is better, but it's complicated by tombstones, although that's not very avoidable, and not exposed anyway. I've put a comment suggesting my current view.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/homestead/465)*

richardwhiuk:
Review status: 3 of 61 files reviewed at latest revision, 34 unresolved discussions, some commit checks failed.

---

*[include/memcached_cache.h, line 24 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krjf5_dHU7FHC3VZeux:-KrmWJM7kboj8NeMVV85:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L24)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Do we need to be passing a pointer by reference?
</blockquote></details>

Done.

---

*[include/memcached_cache.h, line 87 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krjh048VrPJsMacSMGn:-KrmWKrNvP4PbeLROowK:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L87)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

As discussed, should be ADDED not DELETED
</blockquote></details>

Done.

---

*[include/memcached_cache.h, line 181 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkRTLibdna1Z3Feqgb:-KrmWWAy3IQ89U8ZL4u-:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L181)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

It would be good to get a comment explaining what data this represents.
</blockquote></details>

Done.

---

*[include/memcached_cache.h, line 187 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkN-mPfRcx66gS3Kw5:-KrmWXujAUXNF8oj0LU9:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L187)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

This function is just called `get`, which isn't hugely descriptive. A comment/better name would be nice :)
</blockquote></details>

Done.

---

*[include/memcached_cache.h, line 64 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkuH5QnuxHNxihorVN:-KrmWKNIoJeMBp30bypv:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/include/memcached_cache.h#L64)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Also, we need to be setting the XML, Charging Addrs and reigstration state when we create it from an Impu we retrieved from the store
</blockquote></details>

Done.

---

*[src/impu_store.cpp, line 48 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl-ontH_rC5v-_eu6a:-KrmWaNzwLPjEFYrH0pi:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L48)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Did this number come from anywhere in particular?
</blockquote></details>

Done.

---

*[src/impu_store.cpp, line 50 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl0ImiJz31G9vTxG7K:-KrmXGBrQI9ajIsXrl9Z:b-1pyvwk) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L50)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

A comment on this and the decode counterpart explaining why they do this would be good
</blockquote></details>

Done, in the header

---

*[src/impu_store.cpp, line 472 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl6S0Kb2fwOiMP9EFe:-KrmXLJkJo38sy98ny8D:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L472)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Worth a comment pointing out that this one is uncompressed, which is why it's just a JSON string
</blockquote></details>

Done.

---

*[src/impu_store.cpp, line 498 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl6SruMQr2OGR9C4K2:-KrmXM-FyZ58mIm7ow5y:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L498)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

as previous comment
</blockquote></details>

Done.

---

*[src/impu_store.cpp, line 643 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl7_sJJN2CnyRLFIqk:-KrmXNlLU-dquD21BueH:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L643)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Again, I think we need to `try...catch...` this. But can't you use your `extract_array()` helper function you defined above?
</blockquote></details>

Done.

---

*[src/impu_store.cpp, line 664 at r6](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krl8TnbxU5JmO4G7zAJ:-KrmXObAV0k4JtedzLzh:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/10f295d3feeb1cb2ff14418b74907f27b51d4436/src/impu_store.cpp#L664)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Again, may as well use `write_array` from above
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 65 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkQyNrdEvKl4wlnDlf:-KrmXQWaqtcGPdNwruAp:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/src/memcached_cache.cpp#L65)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

This is (necessarily) quite generic, so can we have a comment explaining what it does? I think that `store` will always be a vector of IDs (impis or impus), so renaming that to `store_ids` and `state` to `default_state` might be helpful
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 81 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkS-QYgv7Ufa1_fZ4u:-KrmXRsj0hqNQMa8N1x1:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L81)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Again, a comment. I think I've just figured it out though, so something like:

```
// Updates data with the new elements specified in "updated"
// Elements present in both are left with State `UNCHANGED`
// Elements present in `updated` but not in `data` are set to State `ADDED`
// Elements present in `data` but not in `updated` are set to State `DELETED`
```
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 117 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkT6J1Wfn2G-gYlZMs:-KrmXTQ3RiKCzyIP0Lv-:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L117)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Maybe consider renaming this to `update_from_stored_impu`?
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 139 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkTCC20ZRJSQNpboUX:-KrmXUUzxj-MfX8BIFKU:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L139)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Nice comment :)
Can we get something similar a few lines above, when we're deciding whether to update our member variables with the data from `impu`?
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 246 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkXBBuizYWDYCKG0Go:-KrmXWR3xn67Uc7sE0b8:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L246)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

I can see what these are doing, but a comment explaining why would be nice
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 343 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krkb0SwU0xor6WpN0Hm:-KrmXYLN5Po7gdfiI5Hs:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L343)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Can we also put in a comment here explaining why all these impi mapping changes are using CASed writes, as discussed?
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 539 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krkle5UACiN1i7wlBWV:-KrmXc2kOdoPHUVdZw3s:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L539)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

`get_impu` can return `nullptr` if there's nothing in the store, which I think will happen if we're writing to a remote store for the first time. So presumably we have to protect against this by checking for `nullptr` and creating a new one if so
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 594 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkmMVOojc0Ir-CL3C-:-KrmXd8pcP4AFffrZQ78:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L594)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

May be worth a comment that these three operations happen in sequence, and we can't guarantee consistency between the three tables
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/homestead/465)*

richardwhiuk:
Review status: 3 of 61 files reviewed at latest revision, 34 unresolved discussions, some commit checks failed.

---

*[include/memcached_cache.h, line 102 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krjh9T86I9C8MY1udiX:-KrmXgdJ91-PcINODM_R:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L102)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

As above
</blockquote></details>

Done.

---

*[include/memcached_cache.h, line 162 at r1](https://reviewable.io:443/reviews/metaswitch/homestead/465#-KrkQFUdjmUpdac5dSqQ:-KrmXiIncvMdzmxTTcT8:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/815fd95aa9f2aef560d4dcd484fb3bfcc94887d8/include/memcached_cache.h#L162)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

Maybe update this comment to say that it returns an Impu if the IRS was created from the given store (and hence the CAS value is valid) or `nullptr` if not (ie. if was created for a different store)
</blockquote></details>

Done.

---

*[src/memcached_cache.cpp, line 532 at r5](https://reviewable.io:443/reviews/metaswitch/homestead/465#-Krkkmx-mWoewhn_hZCW:-KrmXlocnfy01KdYf8gS:b-896fix) ([raw file](https://github.com/metaswitch/homestead/blob/7c041a77f59fc0ffcfa9237895feb5c60d383943/src/memcached_cache.cpp#L532)):*
<details><summary><i>Previously, sebrexmetaswitch (Sebastian Rex) wrote…</i></summary><blockquote>

"...or we had a conflict"
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/metaswitch/homestead/465)*

include/homestead_pd_definitions.h
```diff
@@ -106,10 +106,30 @@ static const PDLog2<const char*, int> CL_HOMESTEAD_HTTP_INIT_FAIL
   "(2). Check the network status and configuration. "
 );
 
-static const PDLog CL_HOMESTEAD_ENDED
+static const PDLog CL_HOMESTEAD_CACHE_INIT_FAIL
```
sebrexmetaswitch: In main.cpp, if for some reason we fail to start the HssCacheProcessor

include/hsprov_store.h
```diff
@@ -0,0 +1,132 @@
...
+    /// Get the IMS subscription XML for a public identity.
+    ///
+    /// @param public_id the public identity.
+    GetRegData(const std::string& public_id);
+    virtual ~GetRegData();
+
+    /// Access the result of the request.
+    ///
+    /// @param xml the IMS subscription XML document.
+    virtual void get_xml(std::string& xml);
```
sebrexmetaswitch: these are actually the base classes, and they're virtual to alllow them to be mocked out in UTs

include/hsprov_store.h
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Access the result of the request.
+    ///
+    /// @param charging_addrs the charging addresses for this public identity.
+    virtual void get_charging_addrs(ChargingAddresses& charging_addrs);
```
sebrexmetaswitch: as above

src/diameter_handlers.cpp
```diff
@@ -0,0 +1,465 @@
...
+// completely done with Cache operations.
+// This allows us to delete each of the ImplicitRegistrationSets in the
+// _reg_sets vector in the task's destructor.
+void RegistrationTerminationTask::delete_reg_sets_success()
+{
+  // We have already sent the reponse, so we do nothing here
+  delete this;
+}
+
+// LCOV_EXCL_START - nothing interesting to UT.
```
sebrexmetaswitch: It's easy, just not interesting and that exclusion was there already

include/diameter_hss_connection.h
```diff
@@ -0,0 +1,161 @@
...
+#include "hss_connection.h"
+
+namespace HssConnection {
+
+// Stats the HSS connection can update.
+enum StatsFlags
+{
+  STAT_HSS_LATENCY              = 0x1,
+  STAT_HSS_DIGEST_LATENCY       = 0x2,
+  STAT_HSS_SUBSCRIPTION_LATENCY = 0x4,
```
sebrexmetaswitch: meh, this was existing code so I'm tempted to leave it as-is.

src/diameter_handlers.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+    _cfg->cache->get_implicit_registration_sets_for_impus(success_cb,
+                                                          failure_cb,
+                                                          _impus,
+                                                          this->trail());
+  }
+  else
+  {
+    // This is an invalid deregistration reason.
+    TRC_ERROR("Registration-Termination request received with invalid deregistration reason %d",
```
sebrexmetaswitch: Again, existing code (and we don't necessarily have any impus), so I'm tempted to leave as-is

src/diameter_hss_connection.cpp
```diff
@@ -0,0 +1,504 @@
...
+void DiameterHssConnection::DiameterTransaction<T>::update_latency_stats()
+{
+  if (_stats_manager != NULL)
+  {
+    unsigned long latency = 0;
+    if (get_duration(latency))
+    {
+      if (_stat_updates & STAT_HSS_LATENCY)
+      {
+        _stats_manager->update_H_hss_latency_us(latency);
```
sebrexmetaswitch: I don't know, it was called this already. I think the "us" means "Homestead's view"

src/hsprov_store.cpp
```diff
@@ -0,0 +1,256 @@
...
+      _charging_addrs.ecfs.push_back(it->column.value);
+      TRC_DEBUG("Retrieved secondary_ecf column with value %s",
+                it->column.value.c_str());
+    }
+  }
+
+  // All exceptions will rise up to the calling function, where the return code
+  // will be set as unsuccessful.
+
+  return true;
```
sebrexmetaswitch: Good question. The method is non-void so it overrides the base class' method. As to why we always return true - that's what we did before.

I think the way this works is that the Operation is allowed to verify the data it has got back, and return a boolean success value, so that we can mark cases as failures when we haven't hit an exception

src/hss_cache_processor.cpp
```diff
@@ -0,0 +1,258 @@
...
+void HssCacheProcessor::stop()
+{
+  TRC_STATUS("Stopping threadpool");
+  if (_thread_pool)
+  {
+    _thread_pool->stop();
+  }
+}
+
+void HssCacheProcessor::wait_stopped()
```
sebrexmetaswitch: Now done in the header file

src/hss_cache_processor.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+void HssCacheProcessor::get_implicit_registration_sets_for_impis(irs_vector_success_callback success_cb,
+                                                                 failure_callback failure_cb,
+                                                                 std::vector<std::string> impis,
+                                                                 SAS::TrailId trail)
+{
+  // Create a work item that can run on the thread pool, capturing required
+  // variables to complete the work
+  std::function<void()> work = [this, impis, trail, success_cb, failure_cb]()->void {
```
sebrexmetaswitch: Yeah, but they're capturing different variables etc., so I feel like any attempt to commonalize these would just result in them being completely unreadable. This way at least it's clear what they're doing

src/http_handlers.cpp
```diff
@@ -0,0 +1,1347 @@
...
+  }
+}
+
+std::string ImpuRegDataTask::wildcard_from_body(std::string body)
+{
+  rapidjson::Document document;
+  document.Parse<0>(body.c_str());
+
+  if (!document.IsObject() ||
+      !document.HasMember("wildcard_identity") ||
```
sebrexmetaswitch: maybe, this was just the existing code. Plus makes it easier to add trace I think

src/diameter_handlers.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+    _cfg->cache->get_implicit_registration_sets_for_impis(success_cb,
+                                                          failure_cb,
+                                                          _impis,
+                                                          this->trail());
+  }
+  else if ((!_impus.empty()) && ((_deregistration_reason == PERMANENT_TERMINATION) ||
+                                 (_deregistration_reason == REMOVE_SCSCF)))
+  {
+    // If we have a list of IMPUs to deregister and the reason is allowed, we
```
sebrexmetaswitch: It's checking that _impus isn't empty, hence that we do have a list of IMPUs

include/homestead_pd_definitions.h
```diff
@@ -106,10 +106,30 @@ static const PDLog2<const char*, int> CL_HOMESTEAD_HTTP_INIT_FAIL
...
+  LOG_ERR,
+  "Fatal - No IMPU store specified",
+  "No IMPU store was specified in shared configuration.",
+  "The application will exit and restart until the problem is fixed.",
+  "Specify homestead_impu_store in shared config, and then re-upload shared config."
+);
+
+static const PDLog CL_HOMESTEAD_ENDED
+(
+  PDLogBase::CL_HOMESTEAD_ID + 11,
```
sebrexmetaswitch: I don't think so. I spoke to Rob and he agreed that he thought we could just change this

PR #466, opened 2017-08-09
[No Reviewer] Use common code to calculate diameter timeout

sathiyan-sivathas:
Follow on change from Metaswitch/cpp-common#664.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -265,6 +265,7 @@ int init_options(int argc, char**argv, struct options& options)
 {
   int opt;
   int long_opt_ind;
+  bool diameter_timeout_set = false;
 
   optind = 0;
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -490,6 +492,12 @@ int init_options(int argc, char**argv, struct options& options)
     }
   }
 
+  if (!diameter_timeout_set)
```
eleanor-merry: This makes more sense with the other config validation steps at https://github.com/Metaswitch/homestead/blob/dev/src/main.cpp#L632 surely?

sathiyan-sivathas: I considered this, but at that point I need to make `diameter_timeout_set` global or pass it into `init_options()`, neither of which is very nice. What do you think?

PR #475, opened 2017-09-13
[Reviewer: Richard] Improve cache latency

sebrexmetaswitch:
Cache latency was too high when writing to remote sites because we were performing all the writes synchronously before returning a response to the request that caused us to perform a write.

This resolves that by allowing the `MemcachedCache` to write to the local site first, then call a progress callback (which tells the handler that it has succeeded in writing to the local site), and then continue to write to the remote sites once the handler has processed the local success.

This is done by:
  * passing a `progress_callback` through the `HssCacheProcessor` to the `HssCache` for all put/delete operations
  * changing the `BaseHssCache` to no longer treat a bulk delete as several deletes in a row (since it needs to perform the delete for all IRSs in the local site first, then call the progress callback, then do remote sites)
  * Changing the `MemcachedCache::store_action` to be a `std::function` that captures the data it needs and performs the action
  * using this to allow us to have new actions which delete/put multiple IRSs, used for the RTR/PPR handling
* Changing the handlers' success callbacks to just delete the handler, and moving what was the success callback into a new progress callback

I've confirmed `make full_test` passes, and live tested with 2 sites and confirmed that the remote site writes are happening after the response has been sent.

### DIFF
--- a/include/hss_cache.h
+++ b/include/hss_cache.h
@@ -18,6 +18,7 @@
 #include "ims_subscription.h"
 #include "implicit_reg_set.h"
 
+// The purpose of the progress_callback is explained in hss_cache_processor.h
 typedef std::function<void()> progress_callback;
 
 class HssCache
--- a/include/hss_cache.h
+++ b/include/hss_cache.h
@@ -12,10 +12,13 @@
 #define HSS_CACHE_H_
 
 #include "store.h"
+#include <functional>
 #include <vector>
 #include <string>
 #include "ims_subscription.h"
 #include "implicit_reg_set.h"
+
+typedef std::function<void()> progress_callback;
 
 class HssCache
 {
@@ -51,15 +54,18 @@ public:
   // Save the IRS in the cache
   // Must include updating the impi mapping table if impis have been added
   virtual Store::Status put_implicit_registration_set(ImplicitRegistrationSet* irs,
+                                                      progress_callback progress_cb,
                                                       SAS::TrailId trail) = 0;
 
   // Used for de-registration
   virtual Store::Status delete_implicit_registration_set(ImplicitRegistrationSet* irs,
+                                                         progress_callback progress_cb,
                                                          SAS::TrailId trail) = 0;
 
   // Deletes several registration sets
   // Used for an RTR when we have several registration sets to delete
   virtual Store::Status delete_implicit_registration_sets(const std::vector<ImplicitRegistrationSet*>& irss,
+                                                          progress_callback progress_cb,
                                                           SAS::TrailId trail) = 0;
 
   // Gets the whole IMS subscription for this impi
@@ -71,6 +77,7 @@ public:
 
   // This is used to save the state that we changed in the PPR
   virtual Store::Status put_ims_subscription(ImsSubscription* subscription,
+                                             progress_callback progress_cb,
                                              SAS::TrailId trail) = 0;
 };
 
--- a/src/diameter_handlers.cpp
+++ b/src/diameter_handlers.cpp
@@ -250,10 +250,13 @@ void RegistrationTerminationTask::get_registration_sets_success(std::vector<Impl
     void_success_cb success_cb =
       std::bind(&RegistrationTerminationTask::delete_reg_sets_success, this);
 
+    progress_callback progress_cb =
+      std::bind(&RegistrationTerminationTask::delete_reg_sets_progress, this);
+
     failure_callback failure_cb =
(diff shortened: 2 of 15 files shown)


include/hss_cache.h
```diff
@@ -12,11 +12,14 @@
 #define HSS_CACHE_H_
 
 #include "store.h"
+#include <functional>
 #include <vector>
 #include <string>
 #include "ims_subscription.h"
 #include "implicit_reg_set.h"
 
+typedef std::function<void()> progress_callback;
```
richardwhiuk: Probably worth putting a comment above this describing progress callbacks.

richardwhiuk: Or just pointing to the comment in HssCacheProcessor.

src/diameter_handlers.cpp
```diff
@@ -483,11 +494,16 @@ void PushProfileTask::on_get_ims_sub_failure(Store::Status rc)
...
+void PushProfileTask::on_save_ims_sub_progress()
 {
   SAS::Event event(this->trail(), SASEvent::CACHE_PUT_REG_DATA_SUCCESS, 0);
   SAS::report_event(event);
   send_ppa(DIAMETER_REQ_SUCCESS);
+}
+
+void PushProfileTask::on_save_ims_sub_success()
+{
+  // Just tidy up
```
richardwhiuk: Should we be sending a SAS log when the handler finishes an action?

src/hss_cache_processor.cpp
```diff
@@ -190,14 +192,15 @@ void HssCacheProcessor::delete_implicit_registration_set(void_success_cb success
...
+                                                          progress_callback progress_cb,
                                                           failure_callback failure_cb,
                                                           std::vector<ImplicitRegistrationSet*> irss,
                                                           SAS::TrailId trail)
 {
   // Create a work item that can run on the thread pool, capturing required
   // variables to complete the work
-  std::function<void()> work = [this, irss, trail, success_cb, failure_cb]()->void {
-    Store::Status rc = _cache->delete_implicit_registration_sets(irss, trail);
+  std::function<void()> work = [this, irss, trail, success_cb, progress_cb, failure_cb]()->void {
```
richardwhiuk: Existing, but I think the `{` should be on the next line.

src/hss_cache_processor.cpp
```diff
@@ -239,14 +242,15 @@ void HssCacheProcessor::get_ims_subscription(ims_sub_success_cb success_cb,
...
+                                             progress_callback progress_cb,
                                              failure_callback failure_cb,
                                              ImsSubscription* subscription,
                                              SAS::TrailId trail)
 {
   // Create a work item that can run on the thread pool, capturing required
   // variables to complete the work
-  std::function<void()> work = [this, subscription, trail, success_cb, failure_cb]()->void {
-    Store::Status rc = _cache->put_ims_subscription(subscription, trail);
+  std::function<void()> work = [this, subscription, trail, success_cb, progress_cb, failure_cb]()->void {
```
richardwhiuk: Ditto.

src/memcached_cache.cpp
```diff
@@ -789,16 +844,28 @@ Store::Status MemcachedCache::get_ims_subscription(const std::string& impi,
...
+}
+
+Store::Status MemcachedCache::put_ims_sub_action(ImsSubscription* subscription,
+                                                 SAS::TrailId trail,
+                                                 ImpuStore* store)
+{
   Store::Status status = Store::Status::OK;
 
-  BaseImsSubscription* mis = (BaseImsSubscription*)subscription;
+    BaseImsSubscription* mis = (BaseImsSubscription*)subscription;
```
richardwhiuk: Indentation looks wrong here.

src/memcached_cache.cpp
```diff
@@ -789,16 +844,28 @@ Store::Status MemcachedCache::get_ims_subscription(const std::string& impi,
...
+    for (BaseImsSubscription::Irs::value_type& irs : mis->get_irs())
+    {
+      MemcachedImplicitRegistrationSet* mirs = (MemcachedImplicitRegistrationSet*)irs.second;
+      put_irs_action(mirs, trail, store);
+    }
 
-  return status;
-}
+    return status;
+}
```
richardwhiuk: I think your editor settings are missing the setting to end files in a new line....

src/diameter_handlers.cpp
sebrexmetaswitch: Hmm, possibly, but the memcachedcache will be logging to SAS when it's doing stuff so I'm inclined to leave it as-is.

PR #480, opened 2017-09-21
[Reviewer: AJH] Add watchdog timer config option

tomwatson1024:
Introduces a shared config option to set the delay between watchdog messages in freeDiameter, defaulting to 6s.

The script has undergone very basic live testing, with more to come.

### DIFF
--- a/homestead.root/usr/share/clearwater/infrastructure/scripts/homestead
+++ b/homestead.root/usr/share/clearwater/infrastructure/scripts/homestead
@@ -22,6 +22,12 @@ then
     hs_listen_port=$hss_port
 fi
 
+# If homestead_tw_timer is not set, default to the minimum of 6s
+if [[ -z "$homestead_tw_timer" ]]
+then
+  homestead_tw_timer=6
+fi
+
 # Allow a specific FQDN for Homestead to be set (in case this is an
 # all-in-one node running both ralf and homestead Diameter stacks).
 # Otherwise, fall back to the public hostname.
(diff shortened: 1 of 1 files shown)


homestead.root/usr/share/clearwater/infrastructure/scripts/homestead
```diff
@@ -22,6 +22,12 @@ then
     hs_listen_port=$hss_port
 fi
 
+# If homestead_tw_timer is not set, default to the minimum of 6s
+if [[ -z "$homestead_tw_timer" ]]
+then
+  homestead_tw_timer=6
+fi
```
AlexHockey: A better way to do this is to set the value to 6 unconditionally before the `. /etc/clearwater/config` line. If the config option is specified, it will then overwrite the default value with the configured one.

homestead.root/usr/share/clearwater/infrastructure/scripts/homestead
```diff
@@ -22,6 +22,12 @@ then
     hs_listen_port=$hss_port
 fi
 
+# If homestead_tw_timer is not set, default to the minimum of 6s
+if [[ -z "$homestead_tw_timer" ]]
+then
+  homestead_tw_timer=6
```
AlexHockey: I'm not sure I like the name "*_tw_timer". That won't mean anything to most users I suspect. I would suggest "homestead_diameter_watchdog_timer"

AlexHockey:
:+1:

PR #482, opened 2017-09-22
[Reviewer: Richard] Reinstate the CACHE_PUT_REG_DATA sas log

sebrexmetaswitch:
Also tidy up some unused variables, and fix cases where they were erroneously used

I've tested live that this SAS log appears correctly

### DIFF
--- a/src/http_handlers.cpp
+++ b/src/http_handlers.cpp
@@ -1106,15 +1106,17 @@ void ImpuRegDataTask::put_in_cache()
       }
     }
 
-    SAS::Event event(this->trail(), SASEvent::CACHE_PUT_REG_DATA, 0);
-    std::string public_ids_str = boost::algorithm::join(public_ids, ", ");
-    event.add_var_param(public_ids_str);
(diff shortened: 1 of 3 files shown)


src/http_handlers.cpp
```diff
@@ -1100,12 +1100,22 @@ void ImpuRegDataTask::put_in_cache()
...
         TRC_ERROR("No SIP URI in Implicit Registration Set");
         SAS::Event event(this->trail(), SASEvent::NO_SIP_URI_IN_IRS, 0);
-        event.add_compressed_param(_xml, &SASEvent::PROFILE_SERVICE_PROFILE);
+        event.add_compressed_param(_irs->get_ims_sub_xml(), &SASEvent::PROFILE_SERVICE_PROFILE);
         SAS::report_event(event);
         // LCOV_EXCL_STOP
       }
     }
 
+    SAS::Event event(this->trail(), SASEvent::CACHE_PUT_REG_DATA, 0);
```
richardwhiuk: I think I'd add a block scope here (`{` and `}`), which will mean we scope the variables used to send the SAS event.

Alternatively, I'd have wrapped it up in a function.

PR #484, opened 2017-10-10
[Reviewer: Adam] Allow homestead_http_threads to be configured

richardwhiuk:
Adam,

Hopefully an uncontroversial change to allow homestead's HTTP threads to be configured independently.

I've tested setting `homestead_http_threads` on a Chef deployment and checking that they get picked up.

The equivalent of `sprout_worker_threads`, `homestead_cache_threads` is already standalone.

### DIFF
--- a/debian/homestead.init.d
+++ b/debian/homestead.init.d
@@ -88,6 +88,11 @@ get_settings()
 
         . /etc/clearwater/config
 
+        if [ -n "$homestead_http_threads" ];
+        then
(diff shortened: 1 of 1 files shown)


johadalin:
Sure. 
Though i don't see where the cache_threads are configured separately. They're set to 50x core count here, but thats it. Would it be worth having a similar config flag for them? Most likely i've just missed where we do this already, or its not a relevant thing to do.
Otherwise :+1:

PR #485, opened 2017-10-13
[Reviewer: Richard] Optimize initial registration

MatMeredith:
See comments on https://github.com/Metaswitch/sprout/pull/1965

ToDo: Live test


include/http_handlers.h
```diff
@@ -270,6 +270,12 @@ class ImpuRegDataTask : public HssCacheTask
 
   ImplicitRegistrationSet* _irs = NULL;
 
+  // The last registration state retrieved from the cache for this user.
+  // -- UNKNOWN if we haven't queried the cache
+  // -- UNREGISTERED if we have queried the cache and there was no record
```
richardwhiuk: Should be `NOT_REGISTERED` if we've queried the cache and it wasn't found.

src/homestead_xml_utils.cpp
```diff
@@ -24,17 +24,47 @@ namespace XmlUtils
...
+
+  // If we have been provided with a prev_reg_state then add a
+  // prev_reg_state node.
+  std::string previous_statestring;
+  if (prev_reg_state != RegistrationState::UNKNOWN)
+  {
+    add_reg_state_node(prev_reg_state,
+                       doc,
+                       root,
+                       RegDataXMLUtils::prev_reg_state,
```
richardwhiuk: This looks very inconsistent compared to line 54

include/http_handlers.h
richardwhiuk: Given https://github.com/Metaswitch/homestead/pull/485/files#diff-2212aa57f8d205d68d8b41fd10193c82L836 I think this is what will happen as well.

src/http_handlers.cpp
```diff
@@ -899,12 +899,13 @@ void ImpuRegDataTask::process_received_reg_data()
...
-    // registration or a re-registration. If this subscriber is already
-    // registered but is registering with a new binding, we still need to tell
-    // the HSS.
-    if ((reg_state == RegistrationState::REGISTERED) && (!new_binding))
+    // This message was based on a REGISTER request from Sprout. We need to
+    // send an SAR to the HSS unless ALL of the following are true...
+    // - this is a re-registration
+    // - this is not a new binding
+    // - the request hasn't forbidden us from using cached data
+    // - the existing record hasn't timed out.
```
richardwhiuk: I don't understand why you've changed this comment. It doesn't reflect this if check - it reflects a check further down.

src/ut/http_handlers_test.cpp
```diff
@@ -455,11 +458,14 @@ const std::string HTTPHandlersTest::IMPU4 = "sip:impu4@example.com";
...
-const std::string HTTPHandlersTest::REGDATA_RESULT_INCLUDES_BARRING = "<ClearwaterRegData>\n\t<RegistrationState>REGISTERED</RegistrationState>\n\t<IMSSubscription>\n\t\t<PrivateID>" + IMPI + "</PrivateID>\n\t\t<ServiceProfile>\n\t\t\t<PublicIdentity>\n\t\t\t\t<Identity>" + IMPU + "</Identity>\n\t\t\t\t<BarringIndication>1</BarringIndication>\n\t\t\t</PublicIdentity>\n\t\t\t<PublicIdentity>\n\t\t\t\t<Identity>" + IMPU2 + "</Identity>\n\t\t\t</PublicIdentity>\n\t\t</ServiceProfile>\n\t</IMSSubscription>\n</ClearwaterRegData>\n\n";
+const std::string HTTPHandlersTest::REGDATA_READ_RESULT = "<ClearwaterRegData>\n\t<RegistrationState>REGISTERED</RegistrationState>\n\t<IMSSubscription>\n\t\t<PrivateID>" + IMPI + "</PrivateID>\n\t\t<ServiceProfile>\n\t\t\t<PublicIdentity>\n\t\t\t\t<Identity>" + IMPU + "</Identity>\n\t\t\t</PublicIdentity>\n\t\t\t<PublicIdentity>\n\t\t\t\t<Identity>" + IMPU4 + "</Identity>\n\t\t\t</PublicIdentity>\n\t\t</ServiceProfile>\n\t</IMSSubscription>\n</ClearwaterRegData>\n\n";
```
richardwhiuk: What does READ mean here? Why is this a read result? I think you mean 'Call or Reg'?

docs/homestead_api.md
```diff
@@ -71,7 +71,19 @@ Changes to registration state can be done by:
 
 `PUT /impu/<public ID>/reg-data[?private_id=<private ID>]`
 
-The body of this PUT request is a JSON object, which has a mandatory "reqtype" field, specifying the type of SIP request this state change is based on. The response to a successful PUT request is the same as a GET request (200 OK with an XML body describing the current registration state).
+The body of this PUT request is a JSON object with a mandatory "reqtype" field specifying the type of SIP request this state change is based on. The response to a successful PUT request is the same as a GET request (200 OK with an XML body describing the current registration state), except that for requests of type `reg` or `call` the XML also contains a PreviousRegistrationState element. The PreviousRegistrationState indicates Homestead's local view of the registration state prior to this request (NOT_REGISTERED if Homestead had no cached state for this user). E.g. on an initial registration the response body might be:
```
richardwhiuk: This line is very long - you should be fine to split it just using a single line break.

`E.g. on ...` looks odd to me - I'd prefer `For example, on ...`

src/homestead_xml_utils.cpp
MatMeredith: Argh.   Bad check-in.   It's not inconsistent -- it's just completely broken!

src/http_handlers.cpp
MatMeredith: Because the previous comment was unclear and untrue?   The comment now refers to the whole of the following block of code rather than just the first test, but at least it's accurate?

src/ut/http_handlers_test.cpp
MatMeredith: Because it's used in e.g. ImpuReadRegDataMainline where we are GETting reg-data rather than PUTting it?

src/http_handlers.cpp
richardwhiuk: I think the previous statement was true, but I think your comment is now confusing - with it right next to the if block, it looks like it's describing the if block test, but that's not true.

MatMeredith: I don't really care enough to argue about it :-(   The previous comment wasn't terrible.  I felt I'd improved it slightly! 
 The previous comment said that if this subscriber is already registered but is registering with a new binding then we still need to tell the HSS.   But that's not true.   We _might_ need to tell the HSS in that case.   Whether we _actually_ need to tell the HSS is also dependent on whether we're allowed to use cached data and whether the existing record has timed out.   It seems unnecessarily restrictive to limit comments to only being allowed to describe what's implemented in a single if test -- surely it's often obvious what a single if test is doing, and more helpful to describe what the higher level logic / goal of a chunk of code is?   I agree the comment would be misleading if it said "first check that X, Y, Z and G" if the next line only checked "X and Y".  Oh look -- I seem to  have argued about it anyway :-)

PR #486, opened 2017-10-17
[Reviewer: Adam] Parallelize remote reads of IMPU data

richardwhiuk:
If we fail to read data from the local site, read from all remote sites
simultaneously, and wait for them to return.

We do these reads on a thread pool dedicated to the Memcached Cache.

This means that if we have two sites, which both have a latency of 100ms, we'll only take 200ms to respond, rather than 400ms.

This makes us perform marginally worse in a net split scenario - if both sites have the data, the code will wait for both to return, rather than just asking the first site. If one of the sites has a much lower latency, this will make the performance slightly worse.

I've tested running Clearwater Live Test across a Chef deployment with three sites.

### DIFF
--- a/include/memcached_cache.h
+++ b/include/memcached_cache.h
@@ -15,6 +15,7 @@
 #include "base_ims_subscription.h"
 #include "hss_cache.h"
 #include "impu_store.h"
+#include "threadpool.h"
 
 #include <map>
 #include <string>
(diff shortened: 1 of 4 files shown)


include/memcached_cache.h
```diff
@@ -291,15 +292,29 @@ class MemcachedCache : public BaseHssCache
...
+    _thread_pool.start();
   }
 
   virtual ~MemcachedCache()
   {
+    _thread_pool.stop();
+    _thread_pool.join();
+  }
+
+  // Dummy exception handler callback for the thread pool
```
johadalin: Do we not want a real exception callback? Something that logs out something useful?

src/memcached_cache.cpp
```diff
@@ -248,14 +249,38 @@ ImpuStore::Impu* MemcachedCache::get_impu_for_impu_gr(const std::string& impu,
...
+      _thread_pool.add_work([promise, remote_store, impu, trail]()->void
       {
-        break;
+        promise->set_value(remote_store->get_impu(impu, trail));
+      });
+    }
+
+    for (std::promise<ImpuStore::Impu*>* promise : promises)
+    {
+      std::future<ImpuStore::Impu*> future = promise->get_future();
```
johadalin: Possibly worth a comment stating that this will wait until the future has a valid result, as i had to look it up to check. And that the timeout for that wait will be <thread queue time> + <remote_store get_impu timeout>

include/memcached_cache.h
richardwhiuk: This matches what we do in the cache processor. We don't need to do anything if we've hit an exception, so I think this is fine.

src/memcached_cache.cpp
richardwhiuk: That's just how futures work - I don't think we need to comment that.

PR #490, opened 2017-10-26
[Reviewer: EM] Add grace period to poll_homestead script

tomwatson1024:
This change adds a 20 second grace period after Homestead restarts during which failures of the poll_homestead script will be ignored. This avoids the case where Homestead is repeatedly killed by Monit if the HTTP interface takes a while to come up.

The cost of this change is that it will take Monit 20s longer to kill Homestead if it fails to start.


tomwatson1024:
If the poll_homestead script fails continuously, Monit will restart Homestead once but not again until a success, followed by two successive failures, occurs. I have a fix for this issue - should I include it here?

tomwatson1024:
As discussed, I'm going to look at making some further changes here.

PR #491, opened 2017-10-30
[Reviewer: Richard] Don't blind write UNREGISTERED data to the cache

sebrexmetaswitch:
This is a fix for: https://github.com/Metaswitch/clearwater-issues/issues/2616

As discussed, this changes our writing to the cache so that we don't blind write data with reg-state UNREGISTERED to the cache.

I've put a fairly in-depth comment as to why we're doing this in the code.

I've added a UT for this race window, which fails before the fix and passes after. I've not attempted to replicate the race in live testing, as that didn't seem worth the effort.

At the moment we just ignore DATA_CONTENTION (writing a debug log, but nothing else). Does that sound sensible, or should we be SAS logging?

### DIFF
--- a/src/impu_store.cpp
+++ b/src/impu_store.cpp
@@ -539,8 +539,8 @@ Store::Status ImpuStore::set_impu_without_cas(ImpuStore::Impu* impu,
   return status;
 }
 
-Store::Status ImpuStore::add_impu_without_cas(ImpuStore::Impu* impu,
-                                              SAS::TrailId trail)
+Store::Status ImpuStore::add_impu(ImpuStore::Impu* impu,
+                                  SAS::TrailId trail)
 {
   std::string data;
 
@@ -559,18 +559,9 @@ Store::Status ImpuStore::add_impu_without_cas(ImpuStore::Impu* impu,
                               impu->expiry - now,
                               trail,
                               false);
-
-    if (status == Store::Status::DATA_CONTENTION)
-    {
-      // Just ignore the error
-      TRC_DEBUG("Ignoring data contention error attempting to store IMPU for %s",
-                impu->impu.c_str());
-      status = Store::Status::OK;
-    }
   }
 
   return status;
-
 }
 
 Store::Status ImpuStore::set_impu(ImpuStore::Impu* impu,
--- a/src/impu_store.cpp
+++ b/src/impu_store.cpp
@@ -539,6 +539,38 @@ Store::Status ImpuStore::set_impu_without_cas(ImpuStore::Impu* impu,
   return status;
 }
 
+Store::Status ImpuStore::add_impu_without_cas(ImpuStore::Impu* impu,
+                                              SAS::TrailId trail)
+{
+  std::string data;
+
+  Store::Status status = impu->to_data(data);
+
+  if (status == Store::Status::OK)
+  {
+    int now = time(0);
+
+    // Set the data with a CAS of 0, which will fail if there's data already
+    // present
+    status = _store->set_data("impu",
+                              impu->impu,
+                              data,
+                              0,
+                              impu->expiry - now,
+                              trail,
+                              false);
+
+    if (status == Store::Status::DATA_CONTENTION)
+    {
+      // Just ignore the error
+      status = Store::Status::OK;
+    }
+  }
+
+  return status;
+
+}
+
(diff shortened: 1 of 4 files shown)


src/impu_store.cpp
```diff
@@ -539,6 +539,40 @@ Store::Status ImpuStore::set_impu_without_cas(ImpuStore::Impu* impu,
...
+    {
+      // Just ignore the error
+      TRC_DEBUG("Ignoring data contention error attempting to store IMPU for %s",
+                impu->impu.c_str());
+      status = Store::Status::OK;
+    }
+  }
+
+  return status;
+
```
richardwhiuk: Erroneous blank line.

src/impu_store.cpp
```diff
@@ -539,6 +539,40 @@ Store::Status ImpuStore::set_impu_without_cas(ImpuStore::Impu* impu,
...
+    // present
+    status = _store->set_data("impu",
+                              impu->impu,
+                              data,
+                              0,
+                              impu->expiry - now,
+                              trail,
+                              false);
+
+    if (status == Store::Status::DATA_CONTENTION)
```
richardwhiuk: I don't think we should do this here. I think we should propagate this to the MemcachedCache, which can then make the decision that it's safe to ignore this.

Generally the ImpuStore doesn't have the knowledge to deal with data contention - that's the preserve of the MemcachedCache which has a higher level view.

src/impu_store.cpp
```diff
@@ -539,6 +539,40 @@ Store::Status ImpuStore::set_impu_without_cas(ImpuStore::Impu* impu,
   return status;
 }
 
+Store::Status ImpuStore::add_impu_without_cas(ImpuStore::Impu* impu,
```
richardwhiuk: Possibly just `add_impu`? I don't think `add_impu_with_cas` would be meaningful.

src/memcached_cache.cpp
```diff
@@ -615,13 +615,43 @@ Store::Status MemcachedCache::update_irs_associated_impus(MemcachedImplicitRegis
   return status;
 }
 
+// Create an IMPU from an IRS and set it in the store.
+// This is used when we think that the IMPU is new.
```
richardwhiuk: I'd clarify this and say `This is used when we failed to find the IMPU in the store, and have thus created a new record for it`.

src/memcached_cache.cpp
```diff
@@ -615,13 +615,43 @@ Store::Status MemcachedCache::update_irs_associated_impus(MemcachedImplicitRegis
...
+// The behaviour depends on the registration state of the IMPU we're trying to
+// store:
+//  - if we're storing an IMPU in state REGISTERED, then we just blind write to
+//    the store. This is because we only do this when we've got a new
+//    registration, so we've just communicated with the HSS and the data we have
+//    is authoritative. We therefore want to overwrite anything that's been put
+//    in the store since then
```
richardwhiuk: Missing full stop.

richardwhiuk: Also I'd say:

> This is is safe because either:
> 
> a) the data in the store matches what we have, or is equivalently valid, in which case, what we have can just be updated
> b) the data in the store is from an UNREGISTERED request at the same time. Given that we are about to authoritatively tell Sprout we are registered, we should present a consistent registered state in the store. @
>
> Note, we aren't correcting any differing additional IMPUs, or IMPIs. These should have be subject to a PPR, and the HSS, and will expire anyway. The additional latency in resolving this very narrow race (Multiple SAA and HSS change happening simultaneously to a single IMPU) isn't worthwhile resolving, and it's not obvious which one should win.
```

File: src/memcached_cache.cpp [hunk 6]
```diff
@@ -615,13 +615,43 @@ Store::Status MemcachedCache::update_irs_associated_impus(MemcachedImplicitRegis
...
                                               SAS::TrailId trail,
                                               ImpuStore* store)
 {
   ImpuStore::DefaultImpu* impu = irs->get_impu();
 
-  Store::Status status = store->set_impu_without_cas(impu, trail);
+  Store::Status status = Store::Status::OK;
+
+  if (impu->registration_state == RegistrationState::REGISTERED)
+  {
```
richardwhiuk: Worth a TRC_DEBUG log here?

PR #493, opened 2017-11-03
[Reviewer: Adam] 150 threads per core

sebrexmetaswitch:
Default to 150 homestead cache threads per core, because cache operations are blocking.

I've live tested that this defaults correctly, and we've already live tested that this number of threads fixes the latency issues.


johadalin:
Yup, looks good. :shipit:

PR #496, opened 2017-11-09
[Reviewer: Richard] Don't track cache and HSS latency as part of our HTTP request latency

sebrexmetaswitch:
This PR adds the ability for Homestead to pause the StopWatch on an HTTPRequest when it's performing network I/O.

To achieve this, we've had to pass the StopWatch through the `HssCacheProcessor` and `HssConnection` interfaces.

The Diameter side already tracks Diameter latency, so we just subtract this from our total latency.

The memcached side doesn't, and threading complicates it further. We use the `IOHook`s in cpp-common to register hooks that the `MemcachedStore` will call into when it starts and stops network operations.

These hooks are stored in thread_local data, and are removed when they are deleted. So we must be careful to add the hook before doing any ImpuStore operations, and remove it after we're done.

We also need to be careful not to register any hooks after we've called into any of the `progress_cb` callbacks, as the HttpHandler may have destroyed the StopWatch in response. For this reason, the `perform()` method has been tweaked to only pass the StopWatch to the action for the local store, and not the remotes.

### DIFF
--- a/include/base_hss_cache.h
+++ b/include/base_hss_cache.h
@@ -20,21 +20,25 @@ public:
   // Used for RTR when we have a list of impus
   virtual Store::Status get_implicit_registration_sets_for_impis(const std::vector<std::string>& impis,
                                                                  SAS::TrailId trail,
-                                                                 std::vector<ImplicitRegistrationSet*>& result);
+                                                                 Utils::StopWatch* stopwatch,
+                                                                 std::vector<ImplicitRegistrationSet*>& result) override;
 
   // Get the list of IRSs for the given list of imps
   // Used for RTR when we have a list of impis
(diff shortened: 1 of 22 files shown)


sebrexmetaswitch:
This is not particularly well tested yet, as I've not done any live testing yet and it's not very well unit tested. We could potentially hook or mock or fake stores into `CW_IO_STARTS` and `CW_IO_COMPLETES` to unit test this better.

sebrexmetaswitch:
One other thing that hasn't been added here is any StopWatch manipulation for the HsProvHssConnection. Currently, latency to Cassandra is still included in the request latency.

sebrexmetaswitch:
I've now live-tested and am happy that this is working. You said you might have reviewed this already, but can you check you're happy with this (and the latest commit)

PR #498, opened 2017-11-29
Return large value for timeout if entry not found

Ying-metaswitch:
The table lookup for timeouts may come back as not found for various reasons, e.g. when homestead process is killed. In such case, return a timeout value that's almost guaranteed to be higher than fault tolerance, as it's indicative that all connection attempts may have failed. 

Live tested by stopping homestead and running the script. It will have an error count of 400, and decides "All requests failing? True" if there isn't any successful connection at the same time.

### DIFF
--- a/homestead.root/usr/share/clearwater/bin/check_cx_health.py
+++ b/homestead.root/usr/share/clearwater/bin/check_cx_health.py
@@ -143,8 +143,8 @@ def count_timeouts(table_segment_head):
         _log.debug("Got value {} for timeout OID {}.2.0".format(timeouts[0], table_segment_head))
         return int(timeouts[0])
     else:
-        _log.debug("Timeout OID {}.2.0 entry is empty, possibly homestead process is dead".format(table_segment_head))
         # return a large value as all connections may have timed out.
+        _log.debug("Timeout OID {}.2.0 entry is empty, possibly homestead process is dead".format(table_segment_head))
         return 100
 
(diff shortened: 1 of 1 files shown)


homestead.root/usr/share/clearwater/bin/check_cx_health.py
```diff
@@ -139,9 +139,13 @@ def count_timeouts(table_segment_head):
...
     timeouts = ss.get(netsnmp.VarList(netsnmp.Varbind(table_segment_head + ".2.0")))
-    _log.debug("Got value {} for timeout OID {}.2.0".format(timeouts[0], table_segment_head))
-
-    return int(timeouts[0])
+    if timeouts[0]:
+        _log.debug("Got value {} for timeout OID {}.2.0".format(timeouts[0], table_segment_head))
+        return int(timeouts[0])
+    else:
+        _log.debug("Timeout OID {}.2.0 entry is empty, possibly homestead process is dead".format(table_segment_head))
+        # return a large value as all connections may have timed out.
```
sathiyan-sivathas: Can you put this comment above the log debug statement?

Ying-metaswitch:
Thanks for reviewing! The accumulative error is 400 as `are_all_requests_failing` loops through [10, 11, 12, 13] and each time gets 100 from `count_timeouts`. The default tolerance is 20 in `check_cx_health` script (in the same folder), so I assume 100 is safe enough.

PR #504, opened 2018-01-15
[Reviewer: Richard] Add some virtual destructors where they're missing

sebrexmetaswitch:
As I mentioned on Friday, I spotted some subclasses that didn't have virtual destructors. Figured we should add them so that we don't get tripped up by them in the future. (I've checked that all the classes in `http_handlers.h` and `diameter_handlers.h` now have them.)

### DIFF
--- a/include/diameter_hss_connection.h
+++ b/include/diameter_hss_connection.h
@@ -102,6 +102,8 @@ private:
       _stopwatch(stopwatch)
     {};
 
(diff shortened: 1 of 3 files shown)


sebrexmetaswitch:
I've now also checked the `diameter_hss_connection.h` and `hsprov_hss_connection.h`.

PR #505, opened 2018-01-18
[Reviewer: Adam] Http rework

sebrexmetaswitch:
Update Homestead to use the new HttpConnection interface.

Changes to the product code are:

1. Change the `SproutConnection` to make use of the new interfac
1. Change `main.cpp` to create an `HttpClient` for use by the `SproutConnection`

Changes to the UTs:

1. Split the testing of the `SproutConnection` out from the `DiameterHandlersTest`.
    * instead, `DiameterHandlersTest` just uses a `MockSproutConnection`
    * they're now a bit closer to being actual UTs
    * added a new `sproutconnection_test.cpp`
1. `MockHttpConnection` is now provided by cpp-common, as it should be
1. Remove some unused imports and variables

`make full_test` passes.


src/main.cpp
```diff
@@ -932,12 +932,13 @@ int main(int argc, char**argv)
...
-                                            SASEvent::HttpLogLevel::PROTOCOL,
-                                            NULL);
-  SproutConnection* sprout_conn = new SproutConnection(http);
+  HttpClient* http_client = new HttpClient(false,
+                                           http_resolver,
+                                           SASEvent::HttpLogLevel::PROTOCOL,
+                                           nullptr);
+  HttpConnection* http_conn = new HttpConnection(options.sprout_http_name,
+                                                 http_client);
+  SproutConnection* sprout_conn = new SproutConnection(http_conn);
```
johadalin: Just had to look, as i was going to comment below that we need to delete the httpconnection as well, but a SproutConnection actually takes ownership of the connection and deletes it. Should we be updating it so that is not the case, bringing this more in line with e.g. ChronosConnection? 
Sadly, i've also just looked at hssconnection in Sprout, and that appears to create it's own connection and client, and delete them both as well, so we have no consistency across these connection classes; though it feels like at least having the creator also delete would be a good step forward

src/main.cpp
```diff
@@ -1241,19 +1242,20 @@ int main(int argc, char**argv)
...
+  delete realm_counter; realm_counter = nullptr;
+  delete host_counter; host_counter = nullptr;
+  delete stats_manager; stats_manager = nullptr;
+  delete mar_results_table; mar_results_table = nullptr;
+  delete sar_results_table; sar_results_table = nullptr;
+  delete uar_results_table; uar_results_table = nullptr;
+  delete lir_results_table; lir_results_table = nullptr;
+  delete ppr_results_table; ppr_results_table = nullptr;
+  delete rtr_results_table; rtr_results_table = nullptr;
+  delete cache_queue_size_table; cache_queue_size_table = nullptr;
```
johadalin: 👍 on all this

src/ut/diameter_handlers_test.cpp
```diff
@@ -253,10 +240,12 @@ class DiameterHandlersTest : public testing::Test
           .WillOnce(InvokeArgument<0>(irss));
       }
 
-      // Expect a delete to be sent to Sprout.
-      EXPECT_CALL(*_mock_http_conn, send_delete(http_path, _, body))
-        .Times(1)
-        .WillOnce(Return(http_ret_code)).RetiresOnSaturation();
+      // Expect that we ask the Sprout connection to deregister bindings
+      std::vector<std::string> default_impus = {IMPU3, IMPU};
```
johadalin: Why are the default IMPUs `IMPU3` and `IMPU`? I don't know enough in this area to really follow this. I expect it's all ther right thing to do, but potentially want to check

src/ut/diameter_handlers_test.cpp
```diff
@@ -736,7 +712,8 @@ TEST_F(DiameterHandlersTest, PPRMainline)
     .WillOnce(DoAll(InvokeArgument<1>(), InvokeArgument<0>()));
 
   // And notify Sprout
-  ppr_sprout_connection(IMPU, IMS_SUBSCRIPTION, HTTP_OK);
```
johadalin: Like that we're now actually going to be doing UT here, not using a real sprout connection to generate some body we then expect to be sent.

src/ut/diameter_hss_connection_test.cpp
```diff
@@ -34,19 +34,12 @@
...
 #include <curl/curl.h>
-
 #include "httpstack_utils.h"
-
 #include "mockdiameterstack.hpp"
-#include "mockhttpstack.hpp"
-#include "mockhttpconnection.hpp"
-#include "fakehttpresolver.hpp"
 #include "mockstatisticsmanager.hpp"
-#include "sproutconnection.h"
```
johadalin: Were all of these just unnecessary includes? Impressive

src/ut/sproutconnection_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+  static const std::string change_ids_body;
+
+  SproutConnectionTest()
+  {
+    _mock_http_client = new MockHttpClient();
+    HttpConnection* conn = new HttpConnection("server", _mock_http_client);
+    _sprout_conn = new SproutConnection(conn);
+
+    // If we don't override the default behaviour, return a nonsensical HTTP Code
+    ON_CALL(*_mock_http_client, send_request(_)).WillByDefault(Return(HttpResponse(-1, "", {})));
```
johadalin: Not sure i've put this in everywhere in my tests, which i should check, but should we actually be setting this up in the mock class somehow, rather than expecting users to just know to do this in any new future tests we write?

src/ut/sproutconnection_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  virtual ~SproutConnectionTest()
+  {
+    Mock::VerifyAndClear(_mock_http_client);
```
johadalin: Is this necessary?

src/ut/sproutconnection_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+  MockHttpClient* _mock_http_client;
+  SproutConnection* _sprout_conn;
+};
+
+const std::string SproutConnectionTest::IMPU = "sip:impu@example.com";
+const std::string SproutConnectionTest::IMPI = "_impi@example.com";
+const std::vector<std::string> SproutConnectionTest::IMPIS = { "_impi1@example.com", "_impi2@example.com" };
+const std::vector<std::string> SproutConnectionTest::IMPUS = { "sip:impu1@example.com", "sip:impu2@example.com" };
+const std::string SproutConnectionTest::IMS_SUBSCRIPTION = "<?xml version=\"1.0\"?><IMSSubscription><PrivateID>" + IMPI + "</PrivateID><ServiceProfile><PublicIdentity><Identity>" + IMPU + "</Identity></PublicIdentity></ServiceProfile></IMSSubscription>";
+const std::string SproutConnectionTest::dereg_body = "{\"registrations\":[{\"primary-impu\":\"sip:impu1@example.com\",\"impi\":\"_impi1@example.com\"},{\"primary-impu\":\"sip:impu1@example.com\",\"impi\":\"_impi2@example.com\"},{\"primary-impu\":\"sip:impu2@example.com\",\"impi\":\"_impi1@example.com\"},{\"primary-impu\":\"sip:impu2@example.com\",\"impi\":\"_impi2@example.com\"}]}";
```
johadalin: Don't really care, and i'm not sure splitting it up makes this much nicer, but:  long lines?

src/ut/sproutconnection_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+  // Actually deregister_bindings
+  HTTPCode result = _sprout_conn->deregister_bindings(true, IMPUS, IMPIS, FAKE_TRAIL_ID);
+
+  // Expect that we get the correct return code
+  EXPECT_EQ(HTTP_OK, result);
+}
+
+TEST_F(SproutConnectionTest, DeregisterBindingsWithoutNotifications)
+{
+  // Create a response that will be returned. This one will return an error
```
johadalin: "this one will return an error": returns HTTP_OK
Feels like one of these lines is out of sync

src/ut/sproutconnection_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+const std::string SproutConnectionTest::IMS_SUBSCRIPTION = "<?xml version=\"1.0\"?><IMSSubscription><PrivateID>" + IMPI + "</PrivateID><ServiceProfile><PublicIdentity><Identity>" + IMPU + "</Identity></PublicIdentity></ServiceProfile></IMSSubscription>";
+const std::string SproutConnectionTest::dereg_body = "{\"registrations\":[{\"primary-impu\":\"sip:impu1@example.com\",\"impi\":\"_impi1@example.com\"},{\"primary-impu\":\"sip:impu1@example.com\",\"impi\":\"_impi2@example.com\"},{\"primary-impu\":\"sip:impu2@example.com\",\"impi\":\"_impi1@example.com\"},{\"primary-impu\":\"sip:impu2@example.com\",\"impi\":\"_impi2@example.com\"}]}";
+const std::string SproutConnectionTest::dereg_body_no_impis = "{\"registrations\":[{\"primary-impu\":\"sip:impu1@example.com\"},{\"primary-impu\":\"sip:impu2@example.com\"}]}";
+const std::string SproutConnectionTest::change_ids_body = "{\"user-data-xml\":\"<?xml version=\\\"1.0\\\"?><IMSSubscription><PrivateID>_impi@example.com</PrivateID><ServiceProfile><PublicIdentity><Identity>sip:impu@example.com</Identity></PublicIdentity></ServiceProfile></IMSSubscription>\"}";
```
johadalin: Do we want any comments in this block explaining what these constants are built from? I'm not 100% that i would be able to correctly update these in time, though i'm sure i could work it out. Not sure if there's really a sensible comment we could actually add, but if there is, it would be nice

src/main.cpp
sebrexmetaswitch: Hmm, yeah, I wasn't sure about this at the time, but I think I agree that the fact that the creator doesn't also `delete` it is a bit untidy, as it's not obvious the SproutConnection is taking ownership. I'll change it so SproutConnection doesn't own the HttpConnection

src/ut/diameter_handlers_test.cpp
sebrexmetaswitch: They just are, I guess. The IMS subscription XML specifies that IMPU3 is the default impu of the implicit reg set that contains IMPU2. It's just an arbitrary choice by the author I think

src/ut/sproutconnection_test.cpp
sebrexmetaswitch: I actually quite like that the author of the UTs has to make a conscious decision to add a catch-everything `ON_CALL` method if they want it. Hopefully, it makes people think about whether they really want it, or whether they actually care about the result in every individual UT

src/ut/sproutconnection_test.cpp
sebrexmetaswitch: Nope, because we destroy the mock object just below (and GoogleMock calls this in the destructor anyway). Good spot

src/ut/sproutconnection_test.cpp
sebrexmetaswitch: I've now split these onto multiple lines, which hopefully makes it all a bit clearer. I'll also add a comment

PR #517, opened 2018-03-15
[Reviewer: Seb] Remove reference to retrieving all subscribers

richardwhiuk:
Seb,

As part of the memcached work, we removed the API to list all subscribers which Homestead knows about, as that's not practical on memcached.

This corrects the docs in that area.


sebrexmetaswitch:
LGTM

PR #519, opened 2018-03-16
[Reviewer: Adam] Fix logging

sebrexmetaswitch:
As per https://github.com/Metaswitch/clearwater-issues/issues/3710, a bunch of extra logging added as part of the EC2 testing was accidentally left in. This PR just downgrades/removes those as they're per-call logs.

`make` and `make full_test` both pass.


johadalin:
Taken a quick look over this (was linked from where i was looking), and it looks good. For the cpp-common changes, have we checked all the other things changed in https://github.com/Metaswitch/cpp-common/pull/757/files have been reverted where needed? Looks like your linked change only addresses a single log

sebrexmetaswitch:
Yeah, I searched through all the logs that had been upgraded. I deliverately left the `TRC_STATUS` ones in `realmmanager.cpp`, because I'm those are not per-call, and they're probably useful logs

